PR #1, opened 2017-11-02
RFC: cargo wrapper

dherman:
This RFC proposes we evolve `neon` to be a complete wrapper for `cargo` instead of its own custom abstraction. This also involves a change to the directory structure of Neon projects.

[Rendered](https://github.com/dherman/rfcs-1/blob/master/text/0000-cargo-wrapper.md)

### DIFF
--- a/text/0000-cargo-wrapper.md
+++ b/text/0000-cargo-wrapper.md
@@ -30,7 +30,7 @@ $ neon build
    Compiling regex-syntax v0.4.1
    Compiling unreachable v1.0.0
    Compiling neon v0.1.20
-   Compiling my-neon-lib v0.1.0 (file:///home/dherman/Sources/my-neon-lib/native)
+   Compiling my-neon-lib v0.1.0 (file:///home/dherman/Sources/my-neon-lib)
    Compiling thread_local v0.3.4
    Compiling memchr v1.0.2
    Compiling aho-corasick v0.6.3
@@ -45,15 +45,11 @@ hello neon!
 # Motivation
 [motivation]: #motivation
 
-The purpose of the `neon` CLI tool is to act like cargo for Neon projects: a one-stop workflow tool for generating, building, and managing Neon projects. While abstract over some of the flags required to properly build a Neon project.
-
-## Redesigned `neon` CLI
-
-The long-term goal of the `neon` CLI should be to go away and be replaced by a pure `cargo` plugin, allowing developers to work entirely with a cargo-based workflow. However, this relies on cargo extension hooks that do not yet exist.
+The purpose of the `neon` CLI tool is to act like Cargo for Neon projects: a one-stop workflow tool for generating, building, and managing Neon projects. While abstract over some of the flags required to properly build a Neon project.
 
 ## Modified project layout
 
-The existing Neon project layout is set up so that the top-level directory is an npm package but not a Rust crate, and it _contains_ a Rust crate in the `native/` subdirectory. This means that different tools apply to different parts of the project (`npm` and `neon` to the project root; `cargo` to the `native/` subdirectory). This can also leave the user confused about which directory they should be in and which tool they should use at any given time.
+The current Neon project layout is set up so that the top-level directory is an npm package but not a Rust crate, and it _contains_ a Rust crate in the `native/` subdirectory. This means that different tools apply to different parts of the project (`npm` and `neon` to the project root; `cargo` to the `native/` subdirectory). This can also leave the user confused about which directory they should be in and which tool they should use at any given time.
 
 ```
 my-neon-lib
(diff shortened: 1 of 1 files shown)


text/0000-cargo-wrapper.md
```diff
@@ -0,0 +1,170 @@
...
+│   └── index.js
+├── package.json
+├── README.md
+└── src
+    └── lib.rs
+```
+
+## Redesigned `neon` CLI
+
+The long-term goal of the `neon` CLI should be to go away and be replaced by a pure `cargo` plugin, allowing developers to work entirely with a cargo-based workflow. However, this relies on cargo extension hooks that do not yet exist. So in the meantime, we will provide `neon` as a wrapper around `cargo`, and eventually deprecate this once we can move to `cargo` proper.
```
nixpulvis: Not sure if my comment here is relevant to this line exactly or not, but I'd like to be able to install via `cargo install neon`.

text/0000-cargo-wrapper.md
```diff
@@ -0,0 +1,170 @@
...
+
+### `neon new`
+
+Creates a Neon project, which is always a `--lib` crate. The `--exe` flag is an error.
```
nixpulvis: `--exe` should be `--bin`?

text/0000-cargo-wrapper.md
```diff
@@ -0,0 +1,170 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+A Neon project is both an npm package _and_ a Rust crate, with JS sources in `lib/` and Rust sources in `src/` (following the two ecosystems' conventions). Conceptually, you can think of a Neon project as having a Node _interface_ and a Rust _implementation_. In particular, to the users of the package, they will interact with it as a regular JS library.
+
+Neon provides a modified version of Rust's Cargo command-line tool for creating and working with Neon packages. You can install the Neon command-line interface via:
+
+```
+$ npm install -g neon-cli
```
nixpulvis: I'd like to be able to install via `cargo install neon` as well if possible.

text/0000-cargo-wrapper.md
```diff
@@ -0,0 +1,170 @@
...
+
+The CLI documentation for the `neon` CLI should mostly be taken from the `cargo` CLI documentation, with small modifications according to the changes listed above.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Wrappers are a maintenance burden since they have to keep up with the tool they wrap. They can also lead to confusing or inconsistent error messages. We should pay close attention to error handling to try to make this manageable. This is also somewhat mitigated by explaining `neon` as a thin wrapper, so the user's mental model is "Cargo with the right settings for Neon" instead of a distinct abstraction layer. And of course eventually we want to eliminate this, assuming Cargo will provide the right extension hooks. This seems to be a pretty high priority for the Rust tools subteam so I'm cautiously optimistic.
+
+Some may feel the merged project layout will feel messy or confusing, as opposed to the current cordoning off of the Rust directory under `native/`. However, I think the benefit of decreased decisions and mental context outweighs the messiness, and the most important hygiene constraint--source files kept in separate subdirectories--is still met.
```
nixpulvis: 100% agree.

MilanLoveless:
This PR brings up an existential question that cuts to the heart of "what is Neon?" Should it be low level glue between the run time environments or should it be a high level abstraction that bridges them? Maybe it should be two libraries: the latter built on top of the former. I don't have strong opinions either way, but I am interested in the discussion.

dherman:
@MilanLoveless Can you say a little more about how that question plays out for this RFC? In what ways would you see different answers impacting the design of the tooling and/or project layout?

text/0000-cargo-wrapper.md
dherman: Good catch, thanks! I always mix that up. I’ll fix tomorrow.

nixpulvis:
@MilanLoveless I'm honestly kinda confused the distinction you are making between the two. Maybe an example of each might help?

MilanLoveless:
I guess the clearest example is wrapping Cargo. Do we really want to abstract that for a developer who may or may not be new to writing Rust?

MilanLoveless:
Yeah, that's definitely part of it. Also there is some danger if the development environments (vanilla Rust and Neon) diverge too much.

dherman:
@MilanLoveless Agreed about the divergence concern, which is why I intend this only to be a stopgap until we can actually get the customization hooks we need from cargo. We'd basically be signing up to keep in sync with the cargo behavior. I don't think it changes super frequently, and not only that, a mitigation is the fact that most flags just get passed through untouched. So simple changes as cargo evolves should not require any maintenance work for "nargo."

dherman:
(Cargon?)

nixpulvis:
- `jargo`, feels like Java’s JARs, maybe that’s fine.
- `esargo`, unpronounceable.
- `vargo`, kinda like this one. JS is all about `var` even when it’s `let`.

All I got for you right now.

dherman:
`vargo` is genius! And then you missed an awesome variation on `esargo`... `escargo`!

MilanLoveless:
I like escargo - it's like ECMA Script Cargo 😂

nixpulvis:
Ohhh `escargo` could be good! I love snails ❤️

nixpulvis:
So between `vargo` and `escargo` we have an interesting problem. Vargo sounds more like cargo, but Escargo is more literally correct. Additionally escargo would be easier to “brand” since it’s a type of food. I think I personally like vargo better, but it’s going to make less sense than escargo. 💭

dherman:
I'm not sure "correctness" is exactly an issue with branding, although we do want to create the right mental image for the customer.

I was leaning towards `escargo` until I tried typing both options at the command-line, and `vargo` felt better to me. I think the reason is that `es` is _too_ generic. Neon isn't a general-purpose ECMAScript bridge, it's only for Node. Meanwhile, being a one-letter-change pun on `cargo` is pretty nice, and continues the tradition of `xargo`. And while both names are cute, `vargo` is just kind of amazing. :)

```
$ vargo build
```

nixpulvis:
I completely agree, `vargo` fits into the Rust ecosystem better. And I’m always a fan of less typing.

dherman:
D'oh, I need to close this PR and re-create it from a non-`master` branch. I hope I don't lose the conversation here. :(

dherman:
Phew! Nothing was lost. 😅 I'll link to this thread from the new PR.

nixpulvis:
Haha, I can't count the number of times I've done this.

Issue #6, opened 2017-11-17
Classes 2.0


divmain:
- https://github.com/neon-bindings/neon/issues/19
- https://github.com/neon-bindings/neon/issues/62

dherman:
So the idea in general here is to have a coercion protocol (via traits) and to allow method definitions in classes to have type annotations that use the coercion protocol to abstract away all the conversion boilerplate. Moreover we should make it possible to define functions and modules declaratively with the macro syntax. So at that point we should have a nice "neon mullet" (JS business up front, Rust party in the back). Quick minimal example:

```rust
module! {
    export function foo(x: u32) -> u32 {
        Ok(x + 1)
    }
}
```

nixpulvis:
I'd probably use the word `pub` instead of `export` since we're writing Rust inside `module!`.

dherman:
> I'd probably use the word `pub`

I disagree: the idea is that the outside syntax is JavaScript, and the type syntax and function body syntax are Rust, because it's presenting a JS interface to the outside world while using a Rust implementation (including automatic conversions to Rust types) internally.

matklad:
Some field experience with classes 1.0 (from https://github.com/matklad/fall/). Although the tone will sound slightly negative, I mightily admire the work done, which allows to bind Rust and nodejs at all! :) ❣️ 

Using macro DSL is quite hard: it is opaque, you have to do contortions to impl classes for generic types with `<` in their names, and the only way to abstract over this macro is with another macro. 

Here's how I try to reuse some functionality to declare similar classes: https://github.com/matklad/fall/blob/3d099c94cd3b25ca93e32d39851aaeefab3799ae/code/generic_backend/src/lib.rs#L52-L87. This felt bad to write. 

What's more, this DSL forces huge swaths of Rust code to be inside of macro invocation, which kills a lot of IDE features (of course the IDEs are not mature yet, but it always would be true that the code inside the macro will be harder for IDEs to understand). It also effectively invents a (~~ad-hoc, informally-specified, bug-ridden, slow implementation~~) of new language withing Rust itself. the `method` "keyword" definitely feels alien. 

There are also problems on the client side. I write my client code in TypeScript, and, naturally, I don't have any typechecks when calling Rust classes. 




So, with all that, I find that the best pattern for my use-case was actually C-style OO:

[...]

PR #12, opened 2017-12-08
Add closure support.

abrodersen:
[Rendered](https://github.com/abrodersen/neon-rfcs/blob/closures/text/0000-closures.md)

Add support for Rust closures to the `JsFunction` type.

### DIFF
--- a/text/0000-closures.md
+++ b/text/0000-closures.md
@@ -1,0 +1,85 @@
+- Feature Name: closures
+- Start Date: 2017-12-08
+- RFC PR: (leave this empty)
+- Neon Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Allow `JsFunction` to be initialized from Rust closures, allowing for more advanced interactions between Node.js and Rust.
+
+# Motivation
+[motivation]: #motivation
+
+Currently, Rust functions that can be called from Node.js are very limited. Any data that the function can interact with must be explicitly passed to the function via arguments or `this` binding.
+
+Oftentimes, an API may wich to make use of a continuation passing style to allow asynchronous operations. Consider the following example:
+
+```rust
+doWork(someData, (continue) => {
+  doSomethingAsynchronously(otherData, (result) => { continue(result) });
+});
+```
+
+In this example, the `doWork` function requires a callback parameter. This callback parameter is later invoked, passing another callback named `continue` to be invoked when the sub-operation is complete. A Rust implementation of `continue` cannot have any reference to the context in which it was created. The user of the API would be responsible for passing in all context information explicitly. This would create a clunky and needlessly complex API.
+
+```rust
+// Problem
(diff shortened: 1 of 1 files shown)


dherman:
A very nicely thought-out RFC!

So I think there are some interesting cross-cutting questions with the "classes 2.0" macros (see #6). My vision is for the default and most common idiom for defining both functions and methods is through the front-end macros:

```rust
module! {
    ...
    function increment(x: u32) -> u32 {
        Ok(x + 1)
    }
    ...
}
```

This abstracts away the most annoying boilerplate in Neon, and we should be able to implement it using bare functions instead of closures, to avoid the cost of the extra runtime indirection.

But I also think I agree with the premise of this RFC, which is that in the cases where you're doing it programmatically, you'd like to have the flexibility closures provide available to you.

This makes me think it's actually OK to offer two different constructors for `JsFunction`, one that takes bare functions (call it `JsFunction::from_fn`) and one that takes closures (call it `JsFunction::new`, using your RFC's design). From an API design perspective, I don't usually like to push those kinds of decisions on programmers. But in practice, I think the distribution would be something like 90% of cases use the macro, 9% of cases use the closure API, and 1% of cases use `from_fn`. But it would also allow the macro implementation to use `from_fn` under the hood, to get maximum performance.

abrodersen:
@dherman I think that makes sense. I'll update my pull request to add a `from_fn` constructor for the `JsFunction` struct.

dherman:
I am going to try to see if I can implement this without the `Box` indirection and without rustc getting all riled up. 😛 If we can do that, it's fully backwards-compatible. But either way I'd like to see if I can get this implemented in the 0.2 release.

amckibben:
Any update on this? I'd _really_ love to see closure support.

kjvalencik:
@amckibben I have some concerns about implementing this. There are no stable features in N-API to attach a `Drop` task when the `JsFunction` gets garbage collected. I've also seem some comments from the N-API team that seem to suggest that in `v8` native backed functions _can't_ be garbage collected. This would present a memory leak.

dherman:
> This would present a memory leak.

Unfortunately I think this is going to make this unworkable, as nice as it would be. Neon is just limited by the limitations of V8 and N-API. I filed neon-bindings/neon#518 for us to document the reason for this restriction in the API docs.

ovr:
ref https://github.com/neon-bindings/neon/issues/518#issuecomment-920378958
> Closing because this is something that has since been implemented in Node-API.

Can anyone provide information about this RFC? If it's possible to be implemented via the new NAPI, let's reopen this RFC or I can make a new RFC for this one, because it looks useful.

kjvalencik:
@ovr It might be worth re-writing in the context of the current API which has gone through some changes. Additionally, I don't think this RFC can be implemented as written since it would require GATs (generic associated types).

PR #14, opened 2017-12-20
RFC: VM 2.0

dherman:
Replaces four abstractions of the existing Neon API (`Scope`, `Arguments`, `FunctionCall`, and the `vm` module) with one `Vm` trait, representing a (context-sensitive) view of the JavaScript virtual machine.

[Rendered](https://github.com/neon-bindings/rfcs/blob/2a7bdc26721fedc7b4a25d97a29d67ec898f1fd2/text/0000-vm-2.0.md)

[WIP Implementation](https://github.com/neon-bindings/neon/pull/306)


text/vm-2.0.md
```diff
@@ -0,0 +1,228 @@
...
+A powerful implication of this is that an API that does *not* take a mutable reference to a `Vm` **can never modify the JavaScript virtual machine**! This allows Neon to provide the ability to safely *lock* the virtual machine temporarily, allowing Rust code to inspect and modify the internals of native objects or typed arrays.
+
+## Locking the VM
+
+Some type of JavaScript values—such as typed arrays, or instances of Neon classes—implement an `Inspect` trait, which allows Rust to access their internals. In order to get access to these internals, you first need to *take out a lock* on the JavaScript VM. The `Vm::lock` method “freezes” the VM and produces a temporary `Lock` object, which you can then use to inspect the object.
+
+For example, a simple Neon function that takes a single `ArrayBuffer` object and sets its first byte to 0 would be implemented by calling `vm.lock()` and then `lock.inspect()` with the resulting lock object:
+
+```rust
+fn zero_first_byte(mut vm: impl Vm) -> JsResult<Handle<JsNumber>> {
```
sgrif: Is there a reason that this needs to rely on unstable features? Could this just be `fn zero_first_byte<T: Vm>(mut vm: T)`?

dherman: Oh I was just using the shorthand because it makes me happy :P There's no inherent reliance on it. I should eliminate that to avoid confusion.

text/vm-2.0.md
```diff
@@ -0,0 +1,228 @@
...
+
+These types correspond to different context-sensitive views of the JavaScript virtual machine: the context of a call into a native Rust implementation of a JavaScript function, the context of a call to the `vm.execute()` method, or the context of a call to the `vm.evaluate()` method, respectively. From the user’s perspective, though, the differences between these types are unimportant, and they can simply all be thought of as instances of `Vm`.
+
+## The `Lock` type
+
+The `Lock` type is produced by the `vm.lock()` method described above, and can be used to inspect the internals of JavaScript values:
+
+```rust
+impl Lock {
+    fn inspect<T: Inspect>(&self, value: Handle<T>) -> &mut T::Internals;
```
matklad: This does not look quite right: the `Handle` is copy, so we can get two aliased `&mut T::Internals`?

dherman: Yes, good point. The way the current locking mechanism works is by dynamically tracking what ranges in memory are currently being inspected. We can store that state inside the `Lock` object itself, and either make `inspect` an `&mut self` method or store that state in a cell.

matklad: >The way the current locking mechanism works is by dynamically tracking what ranges in memory are currently being inspected.

Yeah, so I would expect the return type to be `RefMut<T::Internals>`, like `RefCell` works.

dherman: Really interesting point. If we're going all the way to consistency with the stdlib, it makes me wonder if that suggests refactoring the API so that `Inspect` gets a name more like `RefValue` or something like that, with:
```rust
trait RefValue<T> {
    fn borrow(&self, lock: &Lock) -> Ref<T>;
    fn borrow_mut(&mut self, lock: &Lock) -> RefMut<T>;
}
```
So that you would write examples like `zero_first_byte` from the RFC like so:
```rust
fn zero_first_byte(mut vm: impl Vm) -> JsResult<Handle<JsNumber>> {
    let array_buffer = vm.argument(0).unwrap().check::<JsArrayBuffer>()?;
    let lock = vm.lock();
    let mut contents = array_buffer.borrow_mut(&lock);
    let mut buf = contents.as_u8_slice();
    buf[0] = 0;
    Ok(())
}
```
It's appealing to use the nomenclature of borrowing, although since I haven't worked with ref cells in practice, I don't know what kind of learnability hurdles they might introduce. Maybe it's no worse than the learnability of the `Inspect` API though? I really like reusing the intuitions of borrowing for understanding what it means to temporarily inspect the contents of a JS value.

matklad: > If we're going all the way to consistency with the stdlib, 

My point was not about consistency with stdlib (though that'd be nice!). I think that if you do run-time tracking, you'll have to use either closure-base APIs, or return some sort of lock-object, like `RefMut`. Returning plain `&mut T` I think will be unsond.

dherman: @matklad Is there any particular _behavioral_ reason to use `RefMut` instead of `&mut`, or is it just to respect the conventions of the stdlib and send a signal that this is a reference that may dynamically fail to be produced?

matklad: > Is there any particular behavioral reason to use RefMut instead of &mut, or is it just to respect the conventions of the stdlib and send a signal that this is a reference that may dynamically fail to be produced?

I believe there is a behavioral reason. Specifically, I believe that `&mut` returning API will be unsound. I am not 100% sure about this, but looks like what we are doing here is pretty similar to `RefCell`, and it needs those `RefMut` wrappers to be sound.

dherman: Hm, I don't see how to make it work with the actual `Ref`/`RefMut` types of `std::cell`, since those have no constructors and the only way to get one is to call `borrow{_mut}()` on a `RefCell`. I could do something analogous with a custom Neon `Ref` and `RefMut` type though.

dherman: I suppose the soundness issue is needing to track the duration of the borrows? And having a wrapper type allows us to register the borrow on allocation of the Ref{Mut} and unregister it on drop of the Ref{Mut}?

matklad: Yeah, I totally didn't mean that we should use `std::cell::RefMut`, that's impossible. But we quite probably will have to roll our own `neon::RefMut`.

dherman: Got it, thanks! This makes sense to me, and I really like the way the API feels.

matklad: >I suppose the soundness issue is needing to track the duration of the borrows? And having a wrapper type allows us to register the borrow on allocation of the Ref{Mut} and unregister it on drop of the Ref{Mut}?

Exactly, std::RefCell is basically a single-threaded std::Mutex, and std::RefMut is std::LockGuard.

text/vm-2.0.md
```diff
@@ -0,0 +1,338 @@
...
+
+```rust
+struct CallContext<'a, T: This> { /* ... */ };
+
+impl<'a, T: This> CallContext<'a, T> {
+    fn len(&self) -> i32;
+    fn argument_opt(&mut self, i: i32) -> Option<Handle<'a, JsValue>>;
+    fn argument<V: Value>(&mut self, i: i32) -> JsResult<'a, V>;
+    fn this(&mut self) -> Handle<'a, T>;
+    fn callee(&mut self) -> Handle<'a, JsFunction>;
```
tptee: Are we leaving this out of VM 2.0 or removing it later before the 0.2 release?

dherman: Good catch! I'll remove it from the RFC.

dherman:

PR #18, opened 2018-03-22
RFC: Date API

dherman:
A simple API for exposing JS Date objects to Rust.

[Rendered](https://github.com/neon-bindings/rfcs/blob/49f086673a362786431979c3955144fef8b2fe2e/text/0000-date-api.md)

### DIFF
--- a/text/0000-date-api.md
+++ b/text/0000-date-api.md
@@ -1,0 +1,94 @@
+- Feature Name: date_api
+- Start Date: 2018-03-21
+- RFC PR: 
+- Neon Issue: 
+
+# Summary
+[summary]: #summary
+
+This proposal defines an API for creating and inspecting JavaScript `Date` objects.
+
+# Motivation
+[motivation]: #motivation
+
+The JavaScript `Date` API is effectively quite similar to the Rust `std::time::SystemTime` API: a non-monotonic view of the system clock. This type should be reflected into Rust, and it should be possibly to conveniently interoperate between the corresponding JavaScript and Rust APIs.
+
+Example:
+
+```rust
+let date: Handle<JsDate> = JsDate::now(scope)?;
+let time: SystemTime = date.as_time();
+```
+
+# Pedagogy
+[pedagogy]: #pedagogy
+
+This is a relatively straightforward API and can be explained with the API docs.
+
+The API docs should make sure to suggest the [`chrono`](https://github.com/chronotope/chrono) crate as a convenient way to work with date values.
+
+# Details
+[details]: #details
+
+## `JsDate`
+
+This API adds a `JsDate` type to `neon::js`:
+
+```rust
+impl JsDate {
+    pub fn new<'a, S: Scope<'a>, V: AsTimeValue>(_: &mut S, value: V) -> JsResult<JsDate>;
+    pub fn now<'a, S: Scope<'a>>(_: &mut S) -> JsResult<JsDate>;
+    pub fn value(self) -> f64;
+    pub fn as_time(self) -> SystemTime;
+}
+```
+
+The `JsDate::new()` method takes any implementation of `AsTimeValue`, described below.
+
+## `AsTimeValue`
+
+The `AsTimeValue` trait is defined for value types (implementations of `Copy`) that can represent [time values](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-time-values-and-time-range), which can be used to construct `Date` objects.
+
+```rust
+trait AsTimeValue: Copy {
+    fn as_time_value(self) -> f64;
+}
+```
+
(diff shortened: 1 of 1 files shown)


text/0000-date-api.md
```diff
@@ -0,0 +1,60 @@
...
+[details]: #details
+
+## `JsDate`
+
+This API adds a `JsDate` type to `neon::types`:
+
+```rust
+impl JsDate {
+    pub fn new<'a, S: Scope<'a>, V: Into<f64>>(_: &mut S, value: V) -> JsResult<JsDate>;
+    pub fn value(self) -> f64;
```
kjvalencik, line 38: Each of these methods should accept a context for N-API.

text/0000-date-api.md
```diff
@@ -0,0 +1,60 @@
...
+# Details
+[details]: #details
+
+## `JsDate`
+
+This API adds a `JsDate` type to `neon::types`:
+
+```rust
+impl JsDate {
+    pub fn new<'a, S: Scope<'a>, V: Into<f64>>(_: &mut S, value: V) -> JsResult<JsDate>;
```
kjvalencik, line 37: Some `f64` are invalid dates. Should we immediately throw or should we provide an intermediary error that allows the user to handle or call `.or_throw` the user to use `.or_throw`?

dherman: Good question. Constructing an invalid date is not an error in JS:

```text
> var d = new Date(Number.MAX_SAFE_INTEGER + 1000)
undefined
> d
Invalid Date
```

So I think it shouldn't be an error, in order to fully model the semantics of JS.

dherman: Maybe we could also add an `is_valid(&self) -> bool` method.

dherman: The range of [legal date values](https://www.ecma-international.org/ecma-262/11.0/index.html#sec-time-values-and-time-range) is `[-8640000000000000, 8640000000000000]`:

```
> new Date(-8640000000000001)
Invalid Date
> new Date(-8640000000000000)
-271821-04-20T00:00:00.000Z
> new Date(8640000000000000)
+275760-09-13T00:00:00.000Z
> new Date(8640000000000001)
Invalid Date
```

dherman: I added an `is_valid()` method and associated constants for the min and max legal date values.

amilajack:
@kjvalencik @dherman should this be merged if ready?

dherman:
We still have a few open design questions to settle before it's quite ready to merge. Once we do we'll put it into "final comment period" for a week before merging.

text/0000-date-api.md
```diff
@@ -0,0 +1,83 @@
...
+## `JsDate`
+
+This API adds a `JsDate` type to `neon::types`:
+
+```rust
+impl JsDate {
+    pub const MIN_VALID_VALUE: f64 = -8640000000000000;
+    pub const MAX_VALID_VALUE: f64 = 8640000000000000;
+
+    pub fn new<'a, C: Context<'a>, V: Into<f64>>(_: &mut S, value: V) -> JsResult<JsDate>;
```
kjvalencik, line 40: I don't think this should be a `JsResult` since the `Date` API shouldn't throw unless we are already in a throwing state.

I do think it should return an `Error` if the value is out of bounds. I don't think it's safe to depend on being able to get the original value back out since that's not part of the contract/spec.

```rust
impl JsDate {
    /// Produces a _valid_ `Date` or a `DateError`
    pub fn new<'a, C: Context<'a>, V: Into<f64>>(_: &mut S, value: V) -> Result<Handle<'a, JsDate>, DateError>;

    /// Produces a `Date` which _could_ be invalid
    pub fn new_unchecked<'a, C: Context<'a>, V: Into<f64>>(_: &mut S, value: V) -> Handle<'a, JsDate>;
}

enum DateError {
    OutOfRangeLow,
    OutOfRangeHigh,
}

impl Error for DateError {}

impl<'a> JsResultExt<'a, JsString> for Result<Handle<'a, JsDate>, DateError> {
    /// Creates a `RangeError` on error
    fn or_throw<'b, C: Context<'b>>(self, cx: &mut C) -> JsResult<'a, JsDate>;
}
```

text/0000-date-api.md
```diff
@@ -0,0 +1,83 @@
...
+    pub fn value<'a, C: Context<'a>>(self, cx: &mut C) -> f64;
```
kjvalencik, line 41: Note: We should include in the documentation that this _could_ be `NaN`.

amilajack:
Should this be closed or merged? cc @kjvalencik @dherman

kjvalencik:
@amilajack I've added the "final comment period" label. Let's give it a week and if no issues pop-up we can merge. 👍

PR #19, opened 2018-05-19
WIP: an RFC for the pre-build strategy

dherman:
This RFC proposes a strategy for making Neon pre-builds feasible for libraries. This way Neon users can ship libraries whose downstream consumers can be completely unaware of the use of Rust and Neon. In other words, it makes it more tractable to ship native libraries written in Neon with zero requirements for what software their consumers have installed on their machine.

[Rendered](https://github.com/neon-bindings/rfcs/blob/4593c223a36658ff5224256e09e3eb2675058942/text/0000-pre-builds.md)

### DIFF
--- a/text/0000-pre-builds.md
+++ b/text/0000-pre-builds.md
@@ -1,0 +1,53 @@
+- Feature Name: pre_builds
+- Start Date: 2018-05-17
+- RFC PR: 
+- Neon Issue: 
+
+# Summary
+[summary]: #summary
+
+This RFC proposes a strategy for making it practical and feasible for developers of Node libraries implemented with Neon to deploy pre-builds so that downstream consumers can use the library as a dependency _without_ requiring any special software---in particular, the Rust toolchain---installed on their system.
+
+# Motivation
+[motivation]: #motivation
+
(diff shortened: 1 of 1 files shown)


kjvalencik:
NAPI is stable in v10, which becomes LTS this fall. Is this another viable alternative or just something that might ease the implementation?

If there are features in neon that currently cannot be supported by NAPI, what about moving them behind a feature flag?

dherman:
I should say, @matklad made good points, and if the only thing blocking us adopting N-API was that the classes design doesn't fit it, I'm definitely open to revisiting the classes design. The more fundamental feature that we have to support one way or another, though, is the ability to let a JS object own a Rust data structure.

dherman:
OK so the place to look is the [object wrapping](https://nodejs.org/api/n-api.html#n_api_object_wrap) APIs. I'll keep investigating!

tptee:
I think `napi_create_async_work` is the equivalent of `uv_queue_work`? So that’d cover the existing Task implementation using the libuv threadpool. It looks like these functions cover async handles (what’s needed for the Rust thread RFC): https://nodejs.org/api/n-api.html#n_api_custom_asynchronous_operations

dherman:
Here's my proposed course of action: we know we need to explore the N-API backend regardless, and since it sounds like N-API's ABI stability might make for a simpler story than dynamic linking the Neon runtime, we should:
* reject this RFC for now
* start working on an N-API backend
* revisit this RFC's idea only if N-API doesn't solve the problem of large build matrices for some reason

Seem reasonable?

PR #20, opened 2018-06-01
RFC: module re-organization

dherman:
A backwards-incompatible reorganization of the public API organization of `neon::*` to make it easier and more ergonomic to import Neon APIs.

[Rendered](https://github.com/neon-bindings/rfcs/blob/f5b538abc9deb32369342a27524e68224ed1cf72/text/0000-module-reorg.md)

Here's a temporary site where you can [browse the API docs](http://fantastic-detail.surge.sh/neon/index.html).


dherman:
@tptee It just felt like more of a specific use case, and not something you would be likely to want to use in most projects. But we can add it to the prelude compatibly if it becomes clear it's being used most of the time.

dherman:

PR #21, opened 2018-06-18
RFC: infallible `JsString::new()` constructor

dherman:
This draft RFC proposes a backwards-incompatible change to the `JsString::new()` API to make it infallible, along with a `JsString::try_new()` companion API for explicitly handling the (rare) error case of overflowing the maximum string length limit.


dherman:

PR #23, opened 2018-07-09
RFC: error subtyping

dherman:
From the RFC summary:

> To allow for the addition of new error types in the future, as well as to improve the ergonomics of creating and throwing error objects, this RFC replaces the `ErrorKind` enum with a set of new `Js___Error` types that are "subtypes" of `JsError` (using the protocol defined by Neon's `SuperType` trait).


dherman:
📢 This was a bit of a last-minute addition to #22, but I'm going to put it in final comment period as well. We haven't seen much discussion and I'd like to get the 0.2 release out soon. 📢

PR #25, opened 2018-11-30
RFC: EventHandler

geovie:
Provide a way for bindings to call back into JavaScript from threads other than the Node.JS main thread

[Rendered](https://github.com/geovie/rfcs/blob/threadsafecb/text/0000-event-handler.md)

Preliminary implementation: https://github.com/neon-bindings/neon/pull/375

### DIFF
--- a/text/0000-event-handler.md
+++ b/text/0000-event-handler.md
@@ -24,7 +24,7 @@ Being able to call to JavaScript from any thread would be useful for binding aut
 [guide-level-explanation]: #guide-level-explanation
 
 This RFC introduces a new struct `neon::event::EventHandler` that is clone-able and can be send across threads.    
-The user provides a `function`  (and an optional `this` context which defaults to the global object) to the struct constructor.
+The user provides a `function` (and an optional `this` context which defaults to the global object) to the struct constructors.
 To actually invoke JavaScript code the `schedule` method is used. This method accepts a closure which will receive a `neon::Context` and the `this`/`function` pair provided during the creation of `neon::event::EventHandler`.
 
 Example for providing the current progress of a background operation:
@@ -32,21 +32,23 @@ Example for providing the current progress of a background operation:
 ```rust
     let mut this = cx.this();
     let func = cx.argument::<JsFunction>(0)?;
-    let cb = EventHandler::new(this, func);
+    let cb = EventHandler::new(func);
+    // or
+    let cb = EventHandler::bind(this, func);
     thread::spawn(move || {
         for i in 0..100 {
             // do some work ....
             thread::sleep(Duration::from_millis(40));
-            // call into javascript
-            cb.schedule(move |context, this, func| {
-                // call the js function
-                func.call(context, this, vec![cx.number(i).upcast()]);
+            // schedule a call into javascript
+            cb.schedule(move |cx| {
+                // return the arguments of the function call
+                vec![cx.number(i).upcast()]
             }
         }
     });
 ```
 
-Here the `EventHandler` "captures" `this` and `func` and calls the closure from the JavaScript thread with a `context`, the `this` and `func` value.
+Here the `EventHandler` "captures" `this` and `func` and calls the closure from the JavaScript thread with the context (`cx`). The values returned by the closure are used as arguments of the function call.
 
 *Note:* The closure is send to the main Javascript thread so every captured value will be moved.   
 
@@ -74,9 +76,15 @@ impl Drop for EventHandlerInner {
 pub struct EventHandler(Arc<EventHandlerInner>);
 
 impl EventHandler {
-    pub fn new<T: Value>(this: Option<Handle<T>>, function: Handle<JsFunction>) -> Self;
-    pub fn schedule<F>(&self, closure: F)
-        where F: FnOnce(&mut TaskContext, Handle<JsValue>, Handle<JsFunction>),
+    // creates a new EventHandler with global as this
+    pub fn new(function: Handle<JsFunction>) -> Self;
+    // creates a new EventHandler with a custom this
+    pub fn bind<T: Value>(this: Handle<T>, function: Handle<JsFunction>) -> Self;
+
+    // schedule a js function call with the arguments returned from the closure
+    pub fn schedule<T, F>(&self, closure: F)
+        where T: Value
+              F: FnOnce(&mut TaskContext, Handle<JsValue>, Handle<JsFunction>) -> Vec<Handle<T>>,
               F: Send + 'static;
 }
 ```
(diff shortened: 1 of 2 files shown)


amilajack:
@geovie sorry for the delay on this. We'll review this soon

text/0000-threadsafe-callback.md
```diff
@@ -0,0 +1,123 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The main motivation of this is to provide a way to schedule the execution of JavaScript from any (Rust) thread. While `neon::Task` allows to perform a background task and call a JavaScript callback after it's completion, there is currently no way to propagate the progress of a task to JavaScript.   
+Being able to call to JavaScript from any thread would be useful for binding authors trying to bind libraries:
+- with custom threading models
+- with progress state
+- which produce events
+- ...
```
kjvalencik: Also, wrapping C methods which accept a callback.

text/0000-threadsafe-callback.md
```diff
@@ -0,0 +1,123 @@
...
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+There are no real alternatives, calling into JavaScript from any (Rust) thread is a useful feature for bindings author.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Would it make sense to generalize the API e.g. take a `Vec<JsValue>` instead of a `this`/`function` pair?
```
kjvalencik: I don't think so. The only use case I can think of for this is to be able to stash some JS data for easy access when waking up. In practice, I think most of the time you would unwrap this data into native rust types before yielding.

If you must do this, the callback could capture the data and call back into rust, albeit with a small performance cost.

text/0000-threadsafe-callback.md
```diff
@@ -0,0 +1,123 @@
...
+- Is `ThreadSafeCallback` a good name for this? Maybe `ValueCapture`?
```
kjvalencik: I think `ThreadSafe` should be dropped since it is implied by the project. Maybe `vm::Callback` or similar? Could also go with something with `Wake` in the name to patch the `libuv` terminology.

Would `Waker` be too confusing with the new Rust `Waker` API?

dherman: Agreed on `ThreadSafe` being redundant.

I think this API would get clearer if the naming suggested the _role_ rather than the _mechanism_ of the callback. In particular, what you're doing here is creating a callback to serve as an _event handler_, and then at some point later scheduling (from Rust) an _event_ to invoke the handler (from JS).

So I'd suggest calling this API `neon::event::EventHandler`, and instead of a `call` method it should have a `schedule` method. Then the code example from the RFC would look like this:

```rust
let mut this = cx.this();
let func = cx.argument::<JsFunction>(0)?;
let handler = EventHandler::new(this, func);
thread::spawn(move || {
    for i in 0..100 {
        // do some work...
        thread::sleep(Duration::from_millis(40));
        // schedule a javascript event
        handler.schedule(move |cx, this, func| {
            // call the js function
            func.call(cx, this, vec![cx.number(i).upcast()]);
        })
    }
});
```

geovie: @dherman Sounds reasonable to me, I'll try to update the RFC

kjvalencik: @dherman What are your thoughts on this taking a `this` argument? To me, this violates the expectation of typical javascript. I wouldn't expect a callback to have a specified `this` variable unless I specifically bound it.

Is this required to make `() => {}` captures work?

dherman: It seems strictly more expressive to allow it to specify a `this` argument, since you could always choose to use the global object.

Idiomatically, I think at least some event handlers take custom `this` bindings other than the global object. For example: https://nodejs.org/docs/latest/api/events.html#events_passing_arguments_and_this_to_listeners

If it's really uncommon to want to pass a `this` parameter, we could have two different constructors, where the "normal" API defaults to binding to the global object: `EventHandler::new(func)` and a slightly more verbose one lets you specify the `this`-binding: `EventHandler::bound(func, this)`.

> Is this required to make `() => {}` captures work?

Not as I understand it. If `func` is an arrow function, then calling `func.call(cx, this, ...)` will just disregard the `this` argument.

geovie: @dherman No, I just joined. Let's coordinate on slack 👍

text/0000-threadsafe-callback.md
```diff
@@ -0,0 +1,123 @@
...
+- with progress state
+- which produce events
+- ...
+
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This RFC introduces a new struct `neon::ThreadSafeCallback` that is clone-able and can be send across threads.    
+The user provides a `this`/`function` pair to the struct constructor.
```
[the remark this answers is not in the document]
geovie: I changed the RFC so that the `this` object is optional and will default to the `global` object if no `this` is provided

text/0000-threadsafe-callback.md
```diff
@@ -0,0 +1,123 @@
...
+        // free resources
+    }
+}
+
+#[derive(Clone)]
+pub struct ThreadSafeCallback(Arc<ThreadSafeCallbackInner>);
+
+impl ThreadSafeCallback {
+    pub fn new<T: Value>(this: Handle<T>, function: Handle<JsFunction>) -> Self;
+    pub fn call<F>(&self, closure: F)
```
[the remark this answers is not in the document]
geovie: This is tricky and I'm not even sure if it's possible with libuv's API. IMHO the closure is fine.

text/0000-threadsafe-callback.md
geovie: @dherman I've updated the RFC and integrated some of the feedback.

text/0000-event-handler.md
```diff
@@ -0,0 +1,122 @@
...
+    fn drop(&mut self) {
+        // free resources
+    }
+}
+
+#[derive(Clone)]
+pub struct EventHandler(Arc<EventHandlerInner>);
+
+impl EventHandler {
+    pub fn new<T: Value>(this: Option<Handle<T>>, function: Handle<JsFunction>) -> Self;
```
geovie: I made `this` an `Option` for now, not sure if we want to have two constructor functions one which takes `this` and `function` and one which only takes a `function` as argument....

text/0000-threadsafe-callback.md
```diff
@@ -0,0 +1,123 @@
...
+```rust
+    let mut this = cx.this();
+    let func = cx.argument::<JsFunction>(0)?;
+    let cb = ThreadSafeCallback::new(this, func);
+    thread::spawn(move || {
+        for i in 0..100 {
+            // do some work ....
+            thread::sleep(Duration::from_millis(40));
+            // call into javascript
+            cb.call(move |context, this, func| {
```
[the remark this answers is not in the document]
geovie: renamed to `schedule`

text/0000-event-handler.md
kjvalencik: I like the option.

geovie: As discussed on slack I changed this to `EventHandler::new` and `EventHandler::bind`

text/0000-event-handler.md
```diff
@@ -0,0 +1,132 @@
...
+impl EventHandler {
+    // creates a new EventHandler with global as this
+    pub fn new(function: Handle<JsFunction>) -> Self;
+    // creates a new EventHandler with a custom this
+    pub fn bind<T: Value>(this: Handle<T>, function: Handle<JsFunction>) -> Self;
+
+    // schedule a js function call with the arguments returned from the closure
+    pub fn schedule<T, F>(&self, closure: F)
+        where T: Value
+              F: FnOnce(&mut TaskContext, Handle<JsValue>, Handle<JsFunction>) -> Vec<Handle<T>>,
```
geovie: Probably `IntoIterator` can be used instead of `Vec`, but that can be decided during the implementation

z0mbie42:
Hi, I would love to see this completed, merged and released, can I help for anything?

kjvalencik:
@z0mbie42 this is being discussed both in this RFC as well as in the slack workspace. You are welcome to join the conversation if you have thoughts! We don't currently have an expected completion date, but hopefully not long!

text/0000-event-handler.md
kjvalencik: It needs to be turned into a `Vec` for other methods. Making it `IntoIterator` would be inefficient for the common case of already having a vector.

I really like this idea, but I think for it to make sense _everything_ that takes a `Vec<Handle<JsValue>>` should be switched to `IntoIterator`.

kjvalencik:
@dherman This looks good to me. Do you want to make a call for final comment period or given the demand for this feature, should we skip that?

dherman:
I like the simplification and reduction of boilerplate that you get from just being able to compute the arguments list instead of having to explicitly call the handler. In the future if we wanted (say, for performance) a lower-level method that lets you schedule arbitrary logic that take the `this` and `func` as explicit parameters and lets you decide whether to call them:
```rust
cb.schedule_with(move |cx, this, func| { ... })
```
we could add that backwards-compatibly. I don't think it's urgent, though, and I think we should accept the RFC as-is. We can always add more methods later.

Awesome work, and thanks for keeping at it!

dherman:
I've labelled the RFC as being in final comment period! Thanks again @geovie for your work on this.

text/0000-event-handler.md
geovie: @kjvalencik `Vec` already implements `IntoIterator`, so it can be passed directly.

dherman:
🎉🎉🎉🎉🎉 RFC #25 is merged! 🎉🎉🎉🎉🎉

dherman:
RFC #25 is implemented and merged behind a feature flag in neon-bindings/neon#375!

PR #28, opened 2020-05-21
RFC: Exception-safe event handlers

dherman:
This RFC modifies the design of [RFC 25](https://github.com/neon-bindings/rfcs/blob/master/text/0025-event-handler.md) with a solution to handling JavaScript exceptions that might get thrown during the process of launching the event handler.

[Rendered RFC](https://github.com/neon-bindings/rfcs/blob/7cfef089aae2a6527d1032803a8993a0fe2dec63/text/0000-exception-safe-event-handlers.md)

### DIFF
--- a/text/0000-exception-safe-event-handlers.md
+++ b/text/0000-exception-safe-event-handlers.md
@@ -13,40 +13,16 @@
 
 [RFC 25](https://github.com/neon-bindings/rfcs/blob/master/text/0025-event-handler.md) introduced a powerful new API for transmitting a JavaScript event handler to other Rust threads so that they can asynchronously signal events back to the main JavaScript thread.
 
-It offers a low-level API, which gives the Rust thread complete control in how it interacts with the callback:
+However, in that design, the Rust code that prepares the results to send to the event handler has no way to manage operations that can trigger JavaScript exceptions. This shows up even in the simple examples in that RFC, which are forced to use `.unwrap()` to deal with `NeonResult` values:
 
 ```rust
-// Main thread:
-let this = cx.undefined();
-let handler = EventHandler::new(cx, this, f);
+let handler = EventHandler::new(...);
 
-// ...
-
-// Another Rust thread:
-handler.schedule_with(move |cx, this, f| {
-    let args = vec![ /* ... */ ];
-    f.call(cx, this, args).unwrap();
-};
-```
-
-And it offers a high-level API, which only requires the Rust thread to supply a vector of arguments, at which point the API automatically calls the callback:
-
-```rust
-// Main thread:
-let this = cx.undefined();
-let handler = EventHandler::new(cx, this, f);
-
-// ...
-
-// Another Rust thread:
-handler.schedule(move |cx, this, f| {
-    let buf = cx.buffer(1024).unwrap();
-    let args = vec![ buf, /* etc... */ ];
-    args
+handler.schedule(move, |cx, this, f| {
+    let buf = cx.buffer(1024).unwrap(); // panics on exception!
+    ...
 });
 ```
-
-In both cases, there is no good way for code inside the Rust closure to handle operations that might trigger JavaScript exceptions. For example, notice the `unwrap()` calls in both examples above.
 
 For the high-level `schedule()` API, this RFC proposes changing the Rust closure to return a `JsResult` and using the standard, classic Node callback protocol of sending an error value as the first argument (or `null`) and a success value as the second argument (or `null`).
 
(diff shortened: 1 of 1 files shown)


text/0000-exception-safe-event-handlers.md
```diff
@@ -0,0 +1,164 @@
...
+[drawbacks]: #drawbacks
+
+The type signatures are perhaps a bit more complex. But we should lean on the API docs and examples as the way to explain the API, not the type signatures.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+Currently the `neon::event` module is protected by a feature flag, and can't be made generally available until we resolve the problem that it has no way to handle JavaScript exceptions.
+
+We should separately propose a `try_catch()` API for wrapping computations that might throw JavaScript exceptions in a closure and converting the result of the computation in a `Result`. We could leave the `neon::event` API as-is and just tell people to use that. But the high-level API would be less ergonomic, since you'd commonly have to wrap everything in a `try_catch`. Also it seems better to decouple the two APIs and release `neon::event` without having to design and implement a complete `try_catch` solution.
```
kjvalencik, line 159: I think there would be value in implementing this API _using_ `try_catch` instead of in C++, but, that is orthogonal to this RFC.

text/0000-exception-safe-event-handlers.md
```diff
@@ -0,0 +1,164 @@
...
+
+    pub fn new<'a, C: Context<'a>, T: Value>(cx: &C, this: Handle<T>, callback: Handle<JsFunction>) -> Self;
+
+    pub fn schedule<T, F>(&self, arg_cb: F)
+        where T: Value,
+              F: for<'a> FnOnce(&mut EventContext<'a>) -> JsResult<'a, T>,
+              F: Send + 'static;
+
+    pub fn schedule_with<F>(&self, arg_cb: F)
+        where F: FnOnce(&mut EventContext, Handle<JsValue>, Handle<JsFunction>) -> NeonResult<()>,
```
kjvalencik, line 143: Can we make `JsValue` generic so that users don't need to `upcast`?

text/0000-exception-safe-event-handlers.md
```diff
@@ -0,0 +1,164 @@
...
+// Main thread:
+let this = cx.undefined();
+let handler = EventHandler::new(cx, this, f);
+
+// ...
+
+// Another Rust thread:
+handler.schedule_with(move |cx, this, f| {
+    let args = vec![ /* ... */ ];
+    f.call(cx, this, args).unwrap();
```
kjvalencik, line 28: This usage doesn't appear to match the API. Should this changed to...

```rust
handler.schedule_with(move |cx, this, f| {
    let args = vec![ /* ... */ ];
    f.call(cx, this, args)?;

    Ok(())
});
```

dherman: This is describing the API as it appears in RFC 25, but that's not very clear. Let me tighten this section up a bit.

text/0000-exception-safe-event-handlers.md
dherman: The user takes the `JsValue` as an argument, so they don't upcast it. They do have to downcast it if they know it's a particular type but that's not tracked statically. To track it statically we'd need to make `EventHandler` generic in the type of the `this`-binding, maybe something like:

```rust
struct EventHandler<This=JsObject> {
    pub fn new<'a, C: Context<'a>>(cx: &C, this: Handle<This>, callback: Handle<JsFunction<This>>) -> Self;

    pub fn schedule<T, F>(&self, arg_cb: F)
        where T: Value,
              F: for<'a> FnOnce(&mut EventContext<'a>) -> JsResult<'a, T>,
              F: Send + 'static;

    pub fn schedule_with<F>(&self, arg_cb: F)
        where F: FnOnce(&mut EventContext, Handle<This>, Handle<JsFunction<This>>) -> NeonResult<()>,
              F: Send + 'static;

}
```

dherman: Oh, you probably meant for `schedule()` rather than `schedule_with()`!

dherman: Wait, I'm still confused 😂  It looks like `schedule()` already _is_ generic and didn't need the upcast in the example!

text/0000-exception-safe-event-handlers.md
dherman: I'll add a comment about this to this section.

dherman:
This RFC is entering final comment period! If there's no more significant discussion in the next 7 days, it will be accepted and merged!

PR #29, opened 2020-05-30
RFC: try_catch

dherman:
This RFC proposes a new `Context` method for catching any JavaScript exceptions that may be thrown during the execution of a computation. The result of the API uses Rust's standard `Result` enum to distinguish between normal return values and exceptional values:
```rust
match cx.try_catch(|cx| { cx.throw(42) }) {
    Ok(v) => { /* ... computation produced a normal result ... */ }
    Err(v) => { /* ... computation threw an exception ... */ }
}
```

[Rendered](https://github.com/neon-bindings/rfcs/blob/2d4ee792298f1760d5607950ef42b2cd1ca9d482/text/0000-try-catch.md)

### DIFF
--- a/text/0000-try-catch.md
+++ b/text/0000-try-catch.md
@@ -63,24 +63,16 @@ pub trait Context<'a> {
     // N-API only
     fn is_throwing(&self) -> bool;
 
-    fn try_catch<'b: 'a, T, F>(&mut self, f: F) -> Result<Handle<'a, T>, Handle<'a, JsValue>>
+    fn try_catch<T, F>(&mut self, f: F) -> Result<T, Handle<'a, JsValue>>
     where
-        T: Value,
-        F: FnOnce(&mut Self) -> JsResult<'b, T>;
+        T: Sized,
+        F: FnOnce(&mut Self) -> JsResult<'a, T>;
 }
 ```
 
 The `is_throwing` method can be implemented with the N-API [`napi_is_exception_pending`](https://nodejs.org/api/n-api.html#n_api_napi_is_exception_pending) method in the upcoming N-API backend. If the implementation would be too complicated, this method may not be supported with the legacy backend.
 
 The `try_catch` method can be implemented with the N-API [`napi_get_and_clear_last_exception`](https://nodejs.org/api/n-api.html#n_api_napi_get_and_clear_last_exception), or with [`Nan::TryCatch`](https://github.com/nodejs/nan/blob/master/doc/errors.md#api_nan_try_catch) in the legacy backend.
-
-For the legacy backend, the type parameter `F` has an additional constraint:
-
-```rust
-F: std::panic::UnwindSafe
-```
-
-This restriction can be lifted for the N-API backend, which is a backwards-compatible change.
 
 If the callback to `try_catch` produces `Err(Throw)` but the VM is not actually in a throwing state, the `try_catch` function panics.
 
--- a/text/0000-try-catch.md
+++ b/text/0000-try-catch.md
@@ -60,18 +60,29 @@ The `neon::context::Context` trait gets two new methods:
 
 ```rust
 pub trait Context<'a> {
+    // N-API only
     fn is_throwing(&self) -> bool;
-    fn try_catch<T, U, F>(&self, f: F) -> Result<Handle<'a, T>, Handle<'a, U>>
+
+    fn try_catch<'b: 'a, T, F>(&mut self, f: F) -> Result<Handle<'a, T>, Handle<'a, JsValue>>
     where
         T: Value,
-        U: Value,
-        F: for<'b> FnOnce(CatchContext<'b>) -> JsResult<'b, T>;
+        F: FnOnce(&mut Self) -> JsResult<'b, T>;
 }
 ```
 
-The `is_throwing` method can be implemented with the N-API [`napi_is_exception_pending`](https://nodejs.org/api/n-api.html#n_api_napi_is_exception_pending) method in the upcoming N-API backend, or the underlying V8 API in the legacy backend.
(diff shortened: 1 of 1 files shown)


text/0000-try-catch.md
```diff
@@ -0,0 +1,98 @@
...
+The `is_throwing()` method is safe for throwing state.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The `neon::context::Context` trait gets two new methods:
+
+```rust
+pub trait Context<'a> {
+    fn is_throwing(&self) -> bool;
```
kjvalencik, line 63: I don't think this can be easily implemented outside of``N-API`. `N-API` handles this by creating a `v8::TryCatch` each time a function is entered and verifying against that instance.

We may want to consider only supporting this in the `N-API` backend.

text/0000-try-catch.md
```diff
@@ -0,0 +1,98 @@
...
+    fn try_catch<T, U, F>(&self, f: F) -> Result<Handle<'a, T>, Handle<'a, U>>
+    where
+        T: Value,
+        U: Value,
+        F: for<'b> FnOnce(CatchContext<'b>) -> JsResult<'b, T>;
```
kjvalencik, line 68: A try/catch block doesn't change `this` bindings or anything else contextual in Js. Instead of a `CatchContext` should `F` be generic over the context that called `try_catch`?

```rust
pub trait Context<'a> {
    fn try_catch<T, U, F>(&self, f: F) -> Result<Handle<'a, T>, Handle<'a, U>>
    where
        T: Value,
        U: Value,
        F: for<'b> FnOnce(Self) -> JsResult<'b, T>;
}
```

text/0000-try-catch.md
```diff
@@ -0,0 +1,98 @@
...
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The `neon::context::Context` trait gets two new methods:
+
+```rust
+pub trait Context<'a> {
+    fn is_throwing(&self) -> bool;
+    fn try_catch<T, U, F>(&self, f: F) -> Result<Handle<'a, T>, Handle<'a, U>>
```
kjvalencik, line 64: I think this should be `&mut self`.

text/0000-try-catch.md
dherman: We discussed this yesterday on a call—I think it makes sense to keep this method but only offer it for the N-API backend, as you suggest.

dherman: I'll add text to that effect in the RFC.

text/0000-try-catch.md
dherman: Sounds right.

text/0000-try-catch.md
dherman: If we can just eliminate the context and let the closure refer to the outer context, that seems strictly more convenient. It'll probably require implementation to be sure, but for now I'll remove the `CatchContext` type and the context argument.

text/0000-try-catch.md
```diff
@@ -0,0 +1,98 @@
...
+    where
+        T: Value,
+        U: Value,
```
kjvalencik, line 67: After further review, I don't think the error type can be generic. I think it must be `JsValue`. If we were to try to allow it to be generic, it would require a `downcast` to be safe. However, the downcast could fail and we are back to an error that needs to be represented.

kjvalencik:
A user may choose _not_ to operate on the `Result` of a `try_catch`. With the current design, a small amount of boilerplate to transform the `Result<Handle<'a, T>, JsValue<'a>>` into a `JsResult<'a, T>` for returning from a method:

```rust
let value = match result {
    Ok(v) => Ok(v),
    Err(err) => cx.throw(err)?,
};
```

I propose adding to this RFC an implementation of `JsResultExt` to remove ease this boilerplate:

```rust
impl <'a, V, E> JsResultExt<'a, V> for Result<Handle<'a, V>, Handle<'a, E>>
where
    V: Value,
    E: Value,
{
    fn or_throw<'b, C: Context<'b>>(self, cx: &mut C) -> JsResult<'a, V> {
        match self {
            Ok(v) => Ok(v),
            Err(err) => cx.throw(err),
        }
    }
}
```

This would simplify to:

```rust
let value = result.or_throw(value)?;
```

kjvalencik:
Notes from feature flagged implementation:

* There are some additional trait bounds required (notably `UnwindSafe` for legacy)
* Lifetimes are slightly different (requires `'b: 'a`)
* Documentation should include the `panic` behavior on a manually constructed `Err(Throw)`

text/0000-try-catch.md
dherman: We weren't able to eliminate the `cx` argument entirely because of Rust's type rules around `&mut` references. But I did take your suggestion and use `Self` as the type of the inner context and just pass `self` in directly.

text/0000-try-catch.md
```diff
@@ -0,0 +1,112 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+The `neon::context::Context` trait gets two new methods:
+
+```rust
+pub trait Context<'a> {
+    // N-API only
+    fn is_throwing(&self) -> bool;
+
+    fn try_catch<'b: 'a, T, F>(&mut self, f: F) -> Result<Handle<'a, T>, Handle<'a, JsValue>>
```
kjvalencik, line 66: As proposed in `https://github.com/neon-bindings/neon/issues/630`, `try_catch` should return `Result<T, Handle<'a, JsValue>>`. This is strictly more powerful does not require the user to return a `JsValue` which may be unwanted or unnecessary.

Additionally, the lifetime bounds may be more complicated than necessary. They will always be bound to the same `Context`.

```rust
    fn try_catch<T, F>(&mut self, f: F) -> Result<T, Handle<'a, JsValue>> + 'a,
    where
        T: Sized,
        F: FnOnce(&mut Self) -> JsResult<'b, T>;
}
```

kjvalencik (approved):

PR #30, opened 2020-06-01
RFC: TaskBuilder

kjvalencik:
The `Task` API was introduced in Neon `v0.4.0` and has proven to be incredibly valuable to developers. It empowers users to execute long running tasks on the libuv thread pool without blocking the main javascript thread.

The `TaskBuilder` API expands upon this concept with an improved ownership model and ergonomics.

```rust
pub fn add_one(mut cx: FunctionContext) -> JsResult<JsUndefined> {
    let n = cx.argument::<JsNumber>(0)?.value();
    let cb = cx.argument::<JsFunction>(1)?;

    cx.task(move || {
            let result = n + 1.0;

            move |mut cx| Ok(cx.number(result))
        })
        .schedule(cb)
}
```

[Rendered](https://github.com/neon-bindings/rfcs/blob/kv/task-builder/text/0000-task-builder.md)

[Proof of concept implementation](https://github.com/neon-bindings/neon/pull/521)

### DIFF
--- a/text/0000-task-builder.md
+++ b/text/0000-task-builder.md
@@ -20,7 +20,7 @@ While unlikely to cause issues, changing `&self` to `&mut self` is a breaking ch
 Instead of introducing a breaking change, a new API is introduced with an _owned_ `self`. The following addtional changes are made:
 
 * The new API relies on closures instead of a `trait`; thanks to many improvements in the borrow checker and closure ergonomics
-* The new API returns a single `Output` generic rather than separate `Output` and `Error` types. This simplifies infallible tasks.
+* The new API returns a single `Output` generic rather than separate `Output` and `Error` types. This simplifies infallible tasks and is still expressive enough to allow users to use a `Result` type to represent fallible tasks.
 
 # Guide-level explanation
 [guide-level-explanation]: #guide-level-explanation
@@ -32,7 +32,7 @@ The callback follows standard node.js callback style with an error as the first 
 The following example introduces a method named `add_one` which accepts a `Number` and a callback, performing the operation asynchronously.
 
 ```rust
-pub fn add_one(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+pub fn add_one(mut cx: FunctionContext) -> JsResult<JsValue> {
     let n = cx.argument::<JsNumber>(0)?.value();
     let cb = cx.argument::<JsFunction>(1)?;
 
@@ -69,6 +69,16 @@ Any setup that requires the V8 VM (e.g., getting an `f64` from a `Number`) must 
 
 The closure is expected to return _another closure_. This higher-order method will be familiar to many Javascript developers. The returned closure is the `complete` method and will be performed back on the main Javascript thread. In the `complete` closure, the user may convert Rust types back to Javascript types.
 
+The `TaskBuilder::schedule` method submits the task to the queue and also returns an `undefined` to match the idiom of most async Javascript methods.
+
+The returned `JsUndefined` is upcast to `JsValue` and `Ok` wrapped for maximum conveinance, allowing a user to terminate a Neon methon with a call to `.schedule`.
+
+```rust
+impl TaskBuilder {
+    fn schedule(self, callback: Handle<JsFunction>) -> JsResult<JsValue> {}
+}
+```
+
 Finally, a `schedule` method is provided to submit the task to the queue. For conveinance, it returns `JsUndefined`, matching Javascript semantics for most asynchronous methods.
 
 # Reference-level explanation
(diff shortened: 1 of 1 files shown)


qm3ster:
For me, this is more straightforward in terms of "what must I implement?" for a new user, yet terser.
As long as there aren't difficulties with lifetimes (and there shouldn't be) this seems like the way forward.
I do see it as a general quality of life improvement and not directly a solution to https://github.com/neon-bindings/neon/pull/283.

kjvalencik:
@qm3ster I don't think there will be issues with lifetimes in the API itself, but, I do think there's potential to make issues with lifetimes in the user's code more difficult to read.

Error messages on closures can be pretty gnarly, especially when inference failed. For example, I could see a new user failing to `move` the environment on the closure. That type of mistake is more obvious on the `trait` because it's explicit.

qm3ster:
I was accidentally considering someone new to Neon but experienced with Rust closures.
But I definitely see how people just learning Rust to extend their node.js are very likely to be using Neon and how that would be a problem.
My counter is that they would neither be able to call `cx.task` nor implement `Task` by hand without prescriptive documentation at that point, so as long as `move` is accented upon it should be fine, even without going into detail on how closures live in memory.

kjvalencik:
That's an excellent point. New users are likely to copy paste an example. As long as all of our examples include `move` (even where unnecessary--e.g., types implement copy or references are `'static`), we should be in good shape.

If you use `move`, but the field isn't `Send` the error is fairly clear. And to your point, they would encounter the _exact_ same issue with the `Task` trait.

Thanks for the thoughtful response!

text/0000-task-builder.md
```diff
@@ -0,0 +1,200 @@
...
+[motivation]: #motivation
+
+The current `Task` trait passes `self` as a shared reference (`&self`) to `fn perform`. Multiple users (https://github.com/neon-bindings/neon/pull/283) (https://github.com/neon-bindings/neon/pull/520) have found this ownership limiting. Many usages require a mutable or owned `self`. Work-arounds require error prone runtime checks (e.g., `Cell` and `Option`).
+
+While unlikely to cause issues, changing `&self` to `&mut self` is a breaking change. Users could have manually called `fn perform` in a context that relied on shared references.
+
+Instead of introducing a breaking change, a new API is introduced with an _owned_ `self`. The following addtional changes are made:
+
+* The new API relies on closures instead of a `trait`; thanks to many improvements in the borrow checker and closure ergonomics
+* The new API returns a single `Output` generic rather than separate `Output` and `Error` types. This simplifies infallible tasks.
```
dherman, line 23: Mention that it's still more expressive, e.g.: "This simplifies infallible tasks and is still expressive enough to allow users to use a `Result` type to represent fallible tasks."

text/0000-task-builder.md
```diff
@@ -0,0 +1,200 @@
...
+    fn complete<'a>(cx: TaskContext<'a>, result: Self::Output) -> JsResult<Self::JsEvent>;
+    fn schedule(self, callback: Handle<JsFunction>);
+}
+```
+
+### `Task` trait
+
+An implementation of `BaseTask` is added for `T: Task` to provide backwards compatibility to existing implementations of `Task`. Since `complete` is an associated method, `self` is threaded through the `Output` parameter.
+
+The existing `Task` trait will be _deprecated_.
```
dherman, line 109: Let's mention why (it's less convenient and less powerful) and also that we won't remove it until Neon users have had enough time to migrate to the new API.

text/0000-task-builder.md
```diff
@@ -0,0 +1,200 @@
...
+        let (result, task) = output;
+
+        task.complete(cx, result)
+    }
+}
+```
+
+### `TaskBuilder`
+
+The `TaskBuilder` struct contains two pieces of key data:
```
dherman, line 133: Can you include the full `TaskBuilder` type declaration and the signatures of the methods in the RFC?

text/0000-task-builder.md
```diff
@@ -0,0 +1,200 @@
...
+});
+```
+
+The `cx.task(...)` method creates a new `TaskBuilder`. It expects a single argument. This argument is a `closure` that will be executed on the `libuv` threadpool.
+
+Any setup that requires the V8 VM (e.g., getting an `f64` from a `Number`) must be performed outside of this closure.
+
+The closure is expected to return _another closure_. This higher-order method will be familiar to many Javascript developers. The returned closure is the `complete` method and will be performed back on the main Javascript thread. In the `complete` closure, the user may convert Rust types back to Javascript types.
+
+Finally, a `schedule` method is provided to submit the task to the queue. For conveinance, it returns `JsUndefined`, matching Javascript semantics for most asynchronous methods.
```
dherman, line 72: I don't think I understand this. The `schedule` method returns `Handle<JsUndefined>`? Or the JS callback is wrapped to discard its result and return undefined? Or something else? It might be helpful to include the type signature.

text/0000-task-builder.md
```diff
@@ -0,0 +1,200 @@
...
+The `TaskBuilder` struct contains two pieces of key data:
+
+* A mutable reference to `Context`
+    - Allows the conveinance of returning `JsUndefined` from `.schedule`
+    - The `N-API` implementation will require access to `Env`
+* The `task` closure to be performed on another thread
+
+#### Limitation
+
+Ideally we would have a more simple struct type and only call out the specific trait bounds on `fn schedule`.
```
dherman, line 142: I think this section would be easier to understand if we had the real type declaration that this RFC is proposing documented above, to compare and contrast with.

dherman:
@qm3ster I'm curious about this comment:

> not directly a solution to neon-bindings/neon#283.

What is missing from the RFC that doesn't solve those problems? I understood this to be strictly more powerful than an `&mut self`, since it provides an owned `self`.

qm3ster:
@dherman Ah, I just meant that it is more flexible than that, but also would still be boilerplate and readability improvement even if it gave only an immutable reference.

text/0000-task-builder.md
kjvalencik: I'll write more examples. `schedule` returns `undefined` to remove a little more boilerplate of creating that value since that's the most common pattern.

text/0000-task-builder.md
kjvalencik: I'm going to try a little bit more to fix this issue before resigning to the more limited version.

text/0000-task-builder.md
kjvalencik:
```rust
pub fn perform_closure_task(mut cx: FunctionContext) -> JsResult<JsValue> {
    let n = cx.argument::<JsNumber>(0)?.value();
    let cb = cx.argument::<JsFunction>(1)?;

    cx.task(move || {
            let result = n + 1.0;

            move |mut cx| Ok(cx.number(result))
        })
        .schedule(cb)
}
```

kjvalencik: I need to update the RFC, but, I'll made it return a `JsUndefined` upcast to `JsValue`. This makes it fit a few more situations. For example, class methods always return `JsValue`.

dherman:
A thought from a call with @kjvalencik today: can we refactor this API to look a little closer to the [threadsafe handles API](https://github.com/neon-bindings/rfcs/pull/32), for a more consistent overall design?

kjvalencik:
Superseded by https://github.com/neon-bindings/rfcs/pull/35

PR #32, opened 2020-07-28
Threadsafe Handles

kjvalencik:
Two new `Send + Sync` features are introduced: `EventQueue` and `Persistent`. These features work together to allow multi-threaded modules to interact with the Javascript main thread.

* `EventQueue`: Threadsafe handle that allows sending closures to be executed on the main thread.
* `Persistent<T>`: Opaque handle that can be sent across threads, but only dereferenced on the main thread.

### DIFF
--- a/text/0000-thread-safe-handles.md
+++ b/text/0000-thread-safe-handles.md
@@ -1,0 +1,538 @@
+- Feature Name: Threadsafe Handles
+- Start Date: 2020-07-21
+- RFC PR: (leave this empty)
+- Neon Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Two new `Send` and `Sync` features are introduced: `EventQueue: Send + Sync` and `Root: Send`. These features work together to allow multi-threaded modules to interact with JavaScript.
+
+* `EventQueue`: Provides a `send` method for queueing a Rust closure to be executed by the JavaScript thread that created the `EventQueue`.
+* `Root<_>`: An external reference to JavaScript object that can be sent across threads. It may only be dereferenced by the JavaScript thread that created it.
+
+# Motivation
+[motivation]: #motivation
+
+High level event handlers have gone through [several](https://github.com/neon-bindings/rfcs/pull/25) [iterations](https://github.com/neon-bindings/rfcs/pull/28) without quite providing [ideal](https://github.com/neon-bindings/rfcs/issues/31) ergonomics or [safety](https://github.com/neon-bindings/neon/issues/551).
+
+The Threadsafe Handlers feature attempts to decompose the high level `EventHandler` API into lower level, more flexible primitives.
+
+These primitives can be used to build a higher level and _safe_ `EventHandler` API allowing experimentation with patterns outside of `neon`.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Neon provides a smart pointer type `Handle<'a, T: Value>` for referencing values on a JavaScript heap. These references are bound by the lifetime of `Context<'a>` to ensure they can only be used while the VM is locked in a synchronous neon function. These are neither `Send` or `Sync`.
+
+### `neon::handle::Root<T>`
+
+As a developer, I may want to retain a reference to a JavaScript object while returning control back to the VM.
+
+`Handle<'_, _>` may not outlive the context that created them.
+
+```rust
+// Does not compile because `cb` cannot be sent across threads
+fn thread_log(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb = cx.argument::<JsFunction>(0)?;
+
+    std::thread::spawn(move || {
+        println!("{:?}", cb);
+    });
+
+    Ok(cx.undefined())
+}
+```
+
+A `Root<_>` may be created from a `Handle<'_, _>` which can be sent across threads.
+
+```rust
+// Compiles
+fn thread_log(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb = cx.argument::<JsFunction>(0)?.root(&mut cx);
+
+    std::thread::spawn(move || {
+        println!("{:?}", cb);
+    });
+
+    Ok(cx.undefined())
+}
+```
+
+While `Root<_>` may be shared across threads, the inner contents can only be accessed by the JavaScript thread that owns it.
+
+```rust
+fn log(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb_root = cx.argument::<JsFunction>(0)?.root(&mut cx);
+    let cb = cb_root.into_inner(&cx);
+
+    println!("{:?}", cb);
+
+    Ok(cx.undefined())
+}
+```
+
+Our earlier example of `fn thread_log` compiled, but would `panic` when executed. A `Root<_>` may only be dropped on the JavaScript thread that created it. `Root<_>` references that are not unwrapped must be manually dropped.
+
+```rust
+fn log(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb = cx.argument::<JsFunction>(0)?.root(&mut cx);
+
+    // The callback is no longer needed
+    cb.drop(&cx);
+
+    Ok(cx.undefined())
+}
+```
+
+### `neon::handle::EventQueue`
+
+Once a value is wrapped in a `Root<_>`, it must be sent back to the JavaScript that created it to unwrap. `EventQueue` provides a mechanism for requesting work be performed on a JavaScript thread.
+
+To schedule work, send a Rust closure to the event queue:
+
+```rust
+fn thread_callback(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb = cx.argument::<JsFunction>(0)?.root(&mut cx);
+    let queue = cx.event_queue(&mut cx);
+
+    std::thread::spawn(move || {
+        queue.send(move |mut cx| {
+            let this = cx.undefined();
+            let msg = cx.string("Hello, World!");
+            let cb = cb.into_inner(&cx);
+
+            let _ = cb.call(&mut cx, this, vec![msg])?;
+
+            Ok(())
+        });
+    });
+
+    Ok(cx.undefined())
+}
+```
+
+`Root<_>` can be cloned from the JavaScript thread that created them. `EventQueue` cannot be cloned, but are `Sync` and can be wrapped in an `Arc` without addition synchronization.
+
+```rust
+fn thread_callback(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb = cx.argument::<JsFunction>(0)?.root(&mut cx);
+    let queue = Arc::new(cx.event_queue(&mut cx));
+
+    for i in (1..=10) {
+        let queue = queue.clone();
+        let cb = cb.clone(&mut cx);
+
+        std::thread::spawn(move || {
+            queue.send(move |mut cx| {
+                let this = cx.undefined();
+                let msg = cx.string(format!("Count: {}", i));
+                let cb = cb.into_inner(&cx);
+
+                let _ = cb.call(&mut cx, this, vec![msg])?;
+
+                Ok(())
+            });
+        });
+    }
+
+    // Each thread has a clone; drop the original.
+    cb.drop(&cx);
+
+    Ok(cx.undefined())
+}
+```
+
+Instances of `EventQueue` will keep the event loop running and prevent the process from exiting. The `EventQueue::unref` method is provided to change this behavior and allow the process to exit while an instance of `EventQueue` still exists.
+
+However, calls to `EventQueue::send` _might_ not execute before the process exits.
+
+This is identical to the API provided by [timers](https://nodejs.org/api/timers.html) in Node.
+
+```rust
+fn thread_callback(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb = cx.argument::<JsFunction>(0)?.root(&mut cx);
+    let mut queue = cx.event_queue(&mut cx);
+
+    queue.unref();
+
+    std::thread::spawn(move || {
+        std::thread::sleep(std::time::Duration::from_secs(1));
+
+        // If the event queue is empty, the process may exit before this executes
+        queue.send(move |mut cx| {
+            let this = cx.undefined();
+            let msg = cx.string("Hello, World!");
+            let cb = cb.into_inner(&cx);
+
+            let _ = cb.call(&mut cx, this, vec![msg]);
+
+            Ok(())
+        });
+    });
+
+    Ok(cx.undefined())
+}
+```
+
+## Usage with `JsBox`
+
+A user may want to model a stream by storing a callback in a `Root<_>` that can be called many times.
+
+* The `Root<_>` will likely be contained in a `JsBox<_>`
+* `JsBox` requires a `Finalize` trait
+* The `Finalize` trait provides a `finalize` method which will be executed on the JavaScript thread that created the `JsBox`
+
+`Finalize::finalize` can provide an ergonomic way to drop a `Root<_>` when it is contained in a `JsBox<_>` that is no longer needed
+
+```rust
+type MyCounterBox = JsBox<RefCell<MyCounter>>;
+
+struct MyCounter {
+    count: usize,
+    callback: Root<JsFunction>,
+}
+
+impl MyCounter {
+    fn new(callback: Root<JsFunction>) -> Self {
+        Self {
+            count: 0,
+            callback,
+        }
+    }
+
+    fn incr<'a, C: Context<'a>>(&mut self, cx: &mut C) -> JsResult<JsUndefined> {
+        *self.count += 1;
+
+        let this = cx.undefined();
+        let msg = cx.number(self.count as f64);
+        let cb = self.callback.clone(&mut cx).into_inner(&cx);
+
+        let _ = cb.call(&mut cx, this, vec![msg])?;
+
+        Ok(cx.undefined())
+    }
+}
+
+impl Finalize for MyCounter {
+    fn finalize<'a, C: Context<'a>>(self, cx: &mut C) {
+        self.callback.drop(&cx);
+    }
+}
+
+fn my_counter_new(cx: FunctionContext) -> JsResult<MyCounterBox> {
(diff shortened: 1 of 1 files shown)


text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+}
+```
+
+`Persistent<_>` handles that are not unwrapped must be manually dropped. If a `Persistent<_>` is not consumed, neon will `panic` when the value is dropped.
+
+```rust
+fn log(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let persistent_cb = cx.argument::<JsFunction>(0)?.persistent(&mut cx);
+
+    // `Persistent` are reference counted and clone-able.
```
kjvalencik, line 82: This isn't accurate. `clone` would require a `Context`.

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+
+### `neon::handle::EventQueue`
+
+Once a value is wrapped in a `Persistent<_>` handle, it must be sent back to the main JavaScript thread to unwrap. `EventQueue` provides a mechanism for requesting work be peformed on the main thread.
+
+To schedule work, send a closure to the event queue:
+
+```rust
+fn thread_callback(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb = cx.argument::<JsFunction>(0)?.persistent(&mut cx);
```
kjvalencik, line 100: `persistent` isn't specified anywhere.

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- ~Should this be implemented for the legacy runtime or only n-api?~ Only for `n-api`.
+- ~Should `EventQueue` be `Arc` wrapped internally to encourage users to share instances across threads?~ No. This can be covered in documentation and matches the abstraction of neon's proposed [`JsBox`](https://github.com/neon-bindings/rfcs/pull/33) and [`tokio::runtime::Runtime`](https://docs.rs/tokio/0.1.22/tokio/runtime/struct.Runtime.html).
+- ~A global `EventQueue` is necessary for dropping `Persistent`. Should this be exposed?~ No. This is no longer required for the design.
+- ~The global `EventQueue` requires instance data. Are we okay using an [experimental](https://nodejs.org/api/n-api.html#n_api_environment_life_cycle_apis) API?~ This is no longer required for the design.
+- Should `EventQueue::send` accept a closure that returns `()` instead of `NeonResult<()>`? In most cases, the user will want to allow `Throw` to become an `uncaughtException` instead of a `panic` and `NeonResult<()>` provides a small ergonomics improvement.
+- `persistent(&mut cx)` is a little difficult to type. Should it have a less complicated name?
```
kjvalencik, line 362: N-API calls these "references" and the JVM calls them "global references".

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+- Start Date: 2020-07-21
+- RFC PR: (leave this empty)
+- Neon Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Two new `Send` and `Sync` features are introduced: `EventQueue: Send + Sync` and `Persistent: Send`. These features work together to allow multi-threaded modules to interact with the JavaScript main thread.
+
+* `EventQueue`: Threadsafe handle that allows sending closures to be executed on the main thread.
```
dherman, line 11: I'd avoid the word "handle" here to avoid confusion with `Handle`. And maybe be explicit that we're talking about Rust closures.

Also, the RFC talks about "main thread" a lot but it's pretty future-proof to work with multiple JS threads once N-API adds support for workers and postMessage. Seems like we can generally say "JavaScript thread" or "the JavaScript thread the value belongs to" etc instead of "main thread."

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+* `Persistent<T>`: Opaque handle that can be sent across threads, with inner contents only accessible on the main thread.
```
dherman, line 12: Another word for this might be `Shared<T>`. It's a little less cumbersome and daunting, and carries the intuition of shared state.

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+
+A persistent reference may be created from a `Handle<'_, _>` which can be sent across threads.
+
+```rust
+// Compiles
+fn thread_log(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb = cx.argument::<JsFunction>(0)?.persistent(&mut cx);
+
+    std::thread::spawn(move || {
+        // Panics when `cb` is dropped.
```
dherman, line 55: Not critical, but putting this example here in this order is confusing for the reader of the RFC, since you haven't yet explained that persistent handles have to be returned to the JS thread to be properly dropped.

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+        println!("{:?}", cb);
+    });
+
+    cx.undefined()
+}
+```
+
+While `Persistent<_>` may be shared across threads, the inner contents can only be accessed on the main thread.
```
dherman, line 63: This is a place where you could say "on the JS thread" instead of "on the main thread."

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+    // The callback is no longer needed
+    cb.drop(&cx);
+
+    cx.undefined()
+}
+```
+
+### `neon::handle::EventQueue`
+
+Once a value is wrapped in a `Persistent<_>` handle, it must be sent back to the main JavaScript thread to unwrap. `EventQueue` provides a mechanism for requesting work be peformed on the main thread.
```
dherman, line 94: Again: "the main JavaScript thread" and "the main thread" could both be "the JavaScript thread"

dherman: "peformed" => "performed"

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+
+To schedule work, send a closure to the event queue:
+
+```rust
+fn thread_callback(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb = cx.argument::<JsFunction>(0)?.persistent(&mut cx);
+    let queue = EventQueue::new(&mut cx)?;
```
dherman, line 101: I think it would be more idiomatic for constructing an `EventQueue` to be thought of less as an allocation and more as getting a reference to something that's already allocated (and tied to the context). I think the best way to do this would be `cx.event_queue()` instead of `EventQueue::new(&mut cx)`. This also ends up being less awkward since you don't have to explicitly take the `&mut` reference to `cx`.

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+
+            cb.call(&mut cx, this, vec![msg]).unwrap();
+        });
+    });
+
+    cx.undefined()
+}
+```
+
+`Persistent<_>` are clone-able and can be used many times. `EventQueue` are `Sync` and may be shared across threads.
```
dherman, line 117: As @kjvalencik notes above, they can't be clone-able.

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+
+```rust
+fn thread_callback(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let cb = cx.argument::<JsFunction>(0)?.persistent(&mut cx);
```
dherman, line 121: Is there any way we could avoid needing to pass in the context? Since it's coming from a handle, and that handle must have been created in a context, can we not infer the context from the handle?

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+    let queue = Arc::new(EventQueue::new(&mut cx)?);
+
+    let handles = (1..=10).into_iter().map(|i| {
+        let queue = Arc::clone(queue);
```
dherman, line 125: This particular line is giving me some heartburn. Something about it feels particularly difficult to learn and understand. Is there an ambiguity between the `EventQueue` and `Arc` implementations of `Clone` that requires the explicitly namespaced `Arc::clone()` instead of `queue.clone()`? Also why do you need the `Arc` in the first place if the `EventQueue` can be shared across multiple threads already? I'm just having trouble really explaining this part.

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+    // Each thread has a clone; drop the original.
+    cb.drop(&cx);
+
+    cx.undefined()
+}
+```
+
+Instances of `EventQueue` will keep the event loop running and prevent the process from exiting. The `EventQueue::unref` method is provided to change this behavior and allow the process to exit while an instance of `EventQueue` still exists.
+
+However, calls to `EventQueue::schedule` _might_ not execute before the process exits.
```
dherman, line 148: Do you mean they might not execute before the process exits specifically after calling `unref()`? It would be clearer to be explicit here.

dherman: Is `unref()` the name N-API uses? It's a little awkward. I don't feel strongly about it. I wonder if an English verb like `detach()` or `release()` might read better.

dherman: Oh I see, it's paired with `ref()`. Probably ignore this suggestion. :)

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+        &self,
+        f: impl FnOnce(TaskContext) -> NeonResult<()>,
+    ) -> Result<(), EventQueueError>;
+}
+```
+
+The native `napi_call_threadsafe_function` can fail in three ways:
+
+* `napi_queue_full` if the queue is full
+* `napi_invalid_arg` if the thread count is zero.ß This is due to misuse and statically enforced by ownership rules and the `Drop` trait.
```
dherman, line 316: s/ß//

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,362 @@
...
+Both `ref` and `unref` mutate the behavior of the underlying threadsafe function and require exclusive references. These methods are infallible because Rust is capable of statically enforcing the invariants. We may want to optimize with `assert_debug!` on the `napi_status`.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+![standards](https://imgs.xkcd.com/comics/standards.png)
+
+Neon already has `Task`, `EventHandler`, a [proposed](https://github.com/neon-bindings/rfcs/pull/30) `TaskBuilder` and an accepted, but currently unimplemented, [update](https://github.com/neon-bindings/rfcs/pull/28) to `EventHandler`. This is a large amount of API surface area without clear indication of what a user should use.
+
+This can be mitigated with documentation and soft deprecation of existing methods as we get a clearer picture of what a high-level, ergonomic API would look like.
```
dherman, line 347: Worth also clarifying that `Task` and `TaskBuilder` can be collapsed into one thing, and represents access to the Node thread pool, which is orthogonal to sharing handles across threads.

dherman:
Finished my review! This is going to be a great addition to the core primitives of Neon. I'm psyched.

text/0000-thread-safe-handles.md
kjvalencik: I like `Shared`. Do you happen to know if Rust has an analogous type? I.e., a type that can be sent across threads, but only dereferenced on the original thread? It probably doesn't; it doesn't feel very rust-y.

text/0000-thread-safe-handles.md
kjvalencik: @dherman Not without significant refactoring of Neon. Neon requires mutable references to `Context` in many places. We could potentially alias `Env` directly. For what it's worth, I *like* this idea. There are other APIs that could benefit from an internal reference to `Env`, for example, `JsString::value`. However, I would consider that outside of the scope of this RFC since it is consistent with existing API.

text/0000-thread-safe-handles.md
kjvalencik: @dherman `unref` means that this `EventQueue` won't keep the event loop alive. If the event loop otherwise becomes empty and ends, a successfully queued event might not execute. I think `unref` is a good name because it is the idiomatic term used in JavaScript/Node.

https://nodejs.org/api/timers.html#timers_immediate_unref

text/0000-thread-safe-handles.md
kjvalencik: I blame my cat. 🐈  ⌨️

text/0000-thread-safe-handles.md
kjvalencik: `queue` does not implement `clone`. It *could*. This comes down to whether we want to utilize N-APIs underlying atomics on a function or leave it to the user.

The use of `Arc` is orthogonal to it being `Sync`. The `Arc` makes the lifetime `'static`, while `Sync` allows multiple threads to hold references. The `Arc` would be unnecessary here if the example used scoped threads.

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,532 @@
...
+    /// Creates an unbounded queue
+    pub fn new<'a, C: Context<'a>>(cx: &mut C) -> Self;
+
+    /// Allow the Node event loop to exit while this `EventQueue` exists.
+    /// _Idempotent_
+    pub fn unref(&mut self);
+
+    /// Prevent the Node event loop from existing while this `EventQueue` exists. (Default)
+    /// _Idempotent_
+    pub fn reference(&mut self);
```
kjvalencik, line 351: @dherman What do you think of returning `-> &mut self` here to allow for chaining? It doesn't have too much value at the moment, but if we add more mutations it could be useful later without requiring a breaking change.

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,532 @@
...
+let queue = cx.event_queue();
+```
+
+#### Design Notes
+
+The backing `napi_threadsafe_function` is already `Send` and `Sync`, allowing the use of `send` with a shared reference.
+
+Both `ref` and `unref` mutate the behavior of the underlying threadsafe function and require exclusive references. These methods are infallible because Rust is capable of statically enforcing the invariants. We may want to optimize with `assert_debug!` on the `napi_status`.
+
+#### Future Expansion
```
kjvalencik, line 420: In the future, we could make an event queue builder:

```rust
let queue = cx.event_queue_builder()
    .capacity(100)
    .unref()
    .build();
```

text/0000-thread-safe-handles.md
```diff
@@ -0,0 +1,532 @@
...
+The Threadsafe Handlers feature attempts to decompose the high level `EventHandler` API into lower level, more flexible primitives.
+
+These primitives can be used to build a higher level and _safe_ `EventHandler` API allowing experimentation with patterns outside of `neon`.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Neon provides a smart pointer type `Handle<'a, T: Value>` for referencing values on a JavaScript heap. These references are bound by the lifetime of `Context<'a>` to ensure they can only be used while the VM is locked in a synchronous neon function. These are neither `Send` or `Sync`.
+
+### `neon::handle::Root<T>`
```
kjvalencik, line 28: @dherman Do you think these belong in `neon::handle`? I think another option could be `neon::sync` to mirror `std::sync`.

dherman: I think I'd lean towards `neon::handle` because they don't necessarily have to be used for multithreading use cases, right? Like, you can just use them to keep things alive for nonstandard lifetimes.

kjvalencik: Correct. And now that I think about it, calling them `Sync` is really just an implementation detail. Since they can only be read on the JavaScript main thread, they effectively _aren't_ `Sync`. I'll keep this one in `neon::handle`.

I still need a place to put `EventQueue`. Do you have thoughts on that? `sync` is an option, but it doesn't really align with `std::sync` because it's not a type wrapper.

kjvalencik: Maybe `neon::thread`?

dherman: I like `neon::event` since it matches the name, and matches up with the JS concept of event handling and event queues.

dherman: We ended up going with `neon::handle::Root` in the implementation, so I'll close this conversation.

dherman (changes requested):
There's just a couple things out of date now, but otherwise I think this is probably ready for Final Comment Period.

kjvalencik:
@dherman This is close to stabilization. One change I would like to make prior to merging is the addition of a return value. Two APIs to take inspiration from:

* [`std::thread::spawn`](https://doc.rust-lang.org/std/thread/fn.spawn.html)
* [`tokio::spawn`](https://docs.rs/tokio/1.11.0/tokio/fn.spawn.html)

In both of these APIs, a `JoinHandle` is returned. On a `thread::spawn`, a user can call `join` to block until the thread exits and get the return value. Similarly, in `tokio`, the handle can be `await`ed to get the resolved value.

I propose the following changes:

```rust
struct Channel
    pub fn send<T, F>(&self, f: F) -> JoinHandle<T>
    where
        T: Send + 'static,
        F: FnOnce(TaskContext) -> NeonResult<T>;
```

We don't need to _initially_ do anything with the returned `JoinHandle`, but by adding it there, it allows extension without breaking changes to introduce `JoinHandle::join` and `impl Future for JoinHandle`.

Also, I'm still not feeling _certain_ `send` is the right choice. Maybe it should be renamed `spawn`? If it is renamed `spawn`, `Channel` probably isn't the best name either.

dherman:
This is a great idea @kjvalencik.

I think I do like `send` better than `spawn`, though, because we aren't spawning a thread (not even conceptually like a green thread), the user should be thinking of it as sending information back to an existing thread. As far as mixing traditions (channel / join) I think I'm comfortable with it: conceptually we're dealing with a channel because we're communicating between threads, and conceptually we're dealing with joining because we're blocking on another thread's computation.

dherman (approved):
The `JoinHandle` changes look good!

kjvalencik:
Merged in https://github.com/neon-bindings/rfcs/commit/c6396817a56f323ae1d452d373c22513f6d7425b

PR #33, opened 2020-07-30
JsBox RFC

kjvalencik:
`JsBox` provides an opaque smart pointer to data created in Rust and managed by the V8 garbage collector. `JsBox` are a less feature rich, but more ergonomic form of neon [classes](https://docs.rs/neon/0.4.0/neon/macro.declare_types.html).

```rust
fn person_new(mut cx: FunctionContext) -> JsResult<JsBox<Person>> {
    let name = cx.argument::<JsString>(0)?.value(&mut cx);
    let person = Person::new(name);

    Ok(cx.boxed(person))
}

fn person_greet(mut cx: FunctionContext) -> JsResult<JsString> {
    let person = cx.argument::<JsBox<Person>>(0)?.borrow(&mut cx);
    let greeting = person.greet();

    Ok(cx.string(greeting))
}
```

[Rendered RFC](https://github.com/neon-bindings/rfcs/blob/kv/jsbox/text/0000-jsbox.md)


text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+    let id = cx.argument::<JsNumber>(0)?.value();
+    let pool = { /* How do we get a pool? */ };
+
+    let username = pool.get_by_id(id);
+
+    Ok(cx.string(username))
+}
+```
+
+The instance of `Pool` can be returned to Javascript using a `JsBox`. The `Pool` won't be dropped until the `JsBox` is garbage collected.
```
dherman, line 114: JavaScript has a capital S (forgive my pedantry)

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+    /// This should only fail if the VM is in a throwing state or is
+    /// in the process of stopping.
+    pub fn new<'a, C: Context<'a>>(
+        cx: &mut C,
+        v: T,
+    ) -> NeonResult<JsBox>;
+}
+```
+
+`JsBox` can be passed back and forth between Javascript and Rust like any other Js value.
```
dherman, line 180: JavaScript

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+[motivation]: #motivation
+
+Classes require a substantial amount of boilerplate and place a large amount of code inside of a macro. While feature rich, classes can be unergonomic for general use. In many cases, users only need a way to safely pass references across the FFI boundary.
+
+`JsBox` was [first suggested in 2017](https://github.com/neon-bindings/rfcs/issues/6#issuecomment-353403121). There is also evidence that multiple users of Neon only use classes as a thin wrapper, instead relying on glue code in javascript to create classes by composing methods.
+
+Example of boilerplate required to use a Neon Class as a thin wrapper:
+
+```rust
+pub struct Person {
```
dherman, line 21: Formatting nit: seems strange to use tabs or 8 spaces here; Rust convention is 4-space indentation.

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+    let pool = cx.argument::<JsBox<Pool>>(0)?;
+    let id = cx.argument::<JsNumber>(1)?.value();
+
+    let username = pool.get_by_id(id)
+
+    Ok(cx.string(username))
+}
+```
+
+Data may only be borrowed immutability. However, `RefCell` can be used to introduce interior mutability with dynamic borrow checking rules:
```
dherman, line 139: immutably

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+```rust
+fn create_pool(mut cx: FunctionContext) -> JsResult<JsBox<RefCell<Pool>>> {
+    let pool = RefCell::new(Pool::new());
+
+    cx.boxed(pool)
+}
+
+fn set_pool_size(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let size = cx.argument::<JsNumber>(1)?.value() as u32;
+    let mut pool = cx.argument::<JsPool<RefCell<Pool>>>(0)?;
```
dherman, line 150: s/JsPool/JsBox/ here?

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+
+    fn deref(&self) -> &Self::Target;
+}
+```
+
+It is not possible to statically prove that two `JsBox<T>` do not alias the same pointer. For example, if a user passes the same value as two arguments to a function. Therefore, it is only safe for `JsBox<T>` to implement `Deref` and not `DerefMut`.
+
+#### `Send + 'static`
+
+`JsBox` are passed across threads and therefore must be `Send`. It is *not* necessary for `JsBox` to be `Sync` because they may only be borrowed on the main thread.
```
dherman, line 199: Why are `JsBox` passed across threads? I couldn't figure this out from the RFC. For those who aren't fluent in Rust concurrency (I'm, like, intermediate level at best), could you explain a little more about the difference between `Send` and `Sync` and how that applies here?

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+
+##### N-API
+
+* `napi_create_external()`. This API allocates a JavaScript value with external data attached to it.
+* `napi_get_value_external()`. This API retrieves the external data pointer that was previously passed `to napi_create_external()`.
+* `napi_typeof()`. Returns `napi_external` if the type is an external type.
+
+*Note*: Passing an external created by another library, potentially even another neon library, is _undefined behavior_.
+
+Progress is being made to add a [tagging feature](https://github.com/nodejs/node/pull/28237) to more safely unwrap externals. It should be incorporated into future designs; however, it is not included in this proposal.
```
dherman, line 228: This is a little spare on details. Can you explain what extra level of guarantee the tagging feature will provide for the `JsBox` API? What guarantee will we be lacking until we have it available? Or will it only help clean up the implementation?

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+
+##### Rust
+
+```rust
```
dherman, line 232: Can you add a bit of prose here explaining what this section is about?

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+    /// The most common `RefCell` methods are provided on `JsCell` as a
+    /// convenience. Other methods may can be accessed by calling `as_cell`
+    pub fn as_cell<'a, C: Context<'a>>(
+        &self,
+        cx: &mut C,
+    ) -> &RefCell<T>;
+}
+```
+
+This approach has several significant drawbacks:
```
dherman, line 310: This is very compelling!

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+
+Instead, Neon can lean on documentation to guide users towards `RefCell` if they need interior mutability.
+
+### Naming
+
+Several other names were considered:
+
+* `JsExternal`. Matches the name in N-API documentation (external reference); however, `cx.external(v)` might be confusing.
+* `JsRef`. Also, matches N-API but, does not match Rust naming conventions.
+* `JsCell`. Only appropriate for the `RefCell` approach.
```
dherman, line 324: Would it make sense to consider a high-level convenience type, something like `type JsRefCell<T> = JsBox<JsRefCell<T>>;`, for the common case of interior mutability? This would make a common case more concise, e.g. in your `Pool` example above, it would change from
```rust
     let mut pool = cx.argument::<JsBox<RefCell<Pool>>>(0)?;
```
to
```rust
     let mut pool = cx.argument::<JsRefCell<Pool>>(0)?;
```
It would also create an opportunity to document the use case of interior mutability directly in the API docs, making the idiom a little more discoverable.

Maybe not necessary for this RFC, but we could leave a note in "unresolved questions" about future convenience APIs and mention the idea there.

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+
+Overall, `JsBox<T>` behaves very similar to a `std::boxed::Box<T>`.
+
+### Leverage existing borrow API
+
+An existing [`neon::borrow`](https://docs.rs/neon/0.4.0/neon/borrow/index.html) implementation exists. However, the current implementation has multiple soundness holes:
+
+* `Buffer` can be mutably borrowed multiple times
+* `Lock` does not prevent overlapping memory regions
+* Multiple [`Lock`](https://docs.rs/neon/0.4.0/neon/context/struct.Lock.html) can be created allowing aliased mutable borrows
```
dherman, line 334: I'm still kind of shocked about this soundness hole. I could've sworn I checked for it and panic.

Conceptually, I still find the idea of being able to use RAII and Rust mutability rules to lock the main JS thread to be pretty compelling. But I like how you've at least decoupled that idea from the idea of safely exposing Rust data owned by the JS engine back to Rust.

I wonder if we might still have a place for the `Lock` API just as a way to have self-documenting critical sections. OTOH, it might end up being a cute idea that's not particularly useful anymore. Curious what you think?

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should we implement this for the legacy backend? We likely will not since this is a brand new API.
```
dherman, line 339: I say no.

text/0000-jsbox.md
```diff
@@ -0,0 +1,341 @@
...
+- Is it acceptable to let `JsBox::new` panic since it only fails when throwing or shutting down?
```
dherman, line 340: Yes, these are both highly rare cases, which is the spirit of panic.

text/0000-jsbox.md
kjvalencik: It's tabs. I'll switch to spaces because markdown seems to like it better.

With that said, I am a strong supporter of tabs over spaces, regardless of the Rust style guide.

This is an interesting article about why tabs can make your code more accessible and friendly for all developers.
https://dev.to/alexandersandberg/why-we-should-default-to-tabs-instead-of-spaces-for-an-accessible-first-environment-101f

text/0000-jsbox.md
kjvalencik: I'll reword this to emphasize that it's less critical that it's passed across threads and more relevant that ownership is moved.

Most owned types are `Send`. An exception being `Rc`.  The `Rc` docs actually do a pretty good job of explaining what `Send` is and why `Rc` *doesn't* implement it. https://doc.rust-lang.org/stable/std/rc/index.html

The `Send` protects against putting things in a `JsBox` that should leave the current context, even if they implement clone.

text/0000-jsbox.md
kjvalencik: This is directly related to the note above. I'll reformat to make it clearer that they are related.

text/0000-jsbox.md
kjvalencik: I added a section about future extension that might include this. In my opinion, this is above and beyond this RFC since in addition to the type alias we would want other helper methods (e.g., `cx.ref_cell(..)`)

text/0000-jsbox.md
kjvalencik: Currently the RFC defines the APIs as returning a `JsResult`, but that's inconsistent with other methods. I think that should be changed to return an unwrapped `JsBox<_>`.

If we ever decide in the future that we want this potential failure exposed directly, it would be best done all at once for consitency.

text/0000-jsbox.md
kjvalencik: I don't think the `Lock` API fits here because nothing is being locked. I don't understand what you mean by RAII to lock the main js thread. The existing API doesn't take any action to lock the the VM.

kjvalencik: @dherman I thought about this some more and it finally clicked. I think the `Lock` guard API only makes sense for accessing data from `Buffer`.

When accessing a `Buffer`, Neon needs to uphold Rusts guarantees on ownership and mutability of the slice. However, since the data is not owned by Rust, there may exist references unknown to Rust.

For example, imagine a scenario where a user borrows a `&[u8]` to the `Buffer`. According to Rust invariants, this data will not change during the lifetime of the borrow. However, if JavaScript is invoked, e.g. by calling a `JsFunction`, the data could be mutated, violating the invariant.

The `Lock` API solves this by holding an exclusive reference to `Context`. The `Lock` statically prevents calling any Neon methods that could potentially mutate the slice.

This problem does not exist for `JsBox`. The data is owned by Rust and _can't_ be mutated by pure JavaScript. At best, the user could invoke JavaScript that calls another Neon method. However, that Neon method would still be bound by Rusts dynamic borrow checking (`RefCell`). It's simpler, more ergonomic, and no less safe to return shared references without a Lock/Guard on `JsBox`.

Lastly, there _is_ an edge case where the `Lock` API might not uphold invariants. If the `Buffer` was created as an external buffer (memory created and managed externally), it's possible for other non-Neon code to mutate these types of buffers. Neon could use `n-api` methods for detecting externals and refuse to lend them. Additionally, neon could provide `unsafe fn borrow_unchecked` methods for reading data from externals.

dherman: It took me a couple of reads but I think you've really nailed the difference:

- Rust data (`JsBox`) **can** be protected by Rust's `Cell` invariants, so we **don't** need to worry about reborrows via VM re-entrancy;
- C/C++ data (`JsArrayBuffer`) **can't** be protected by Rust's `Cell` invariants, so we **do** need to worry about reborrows via VM re-entrancy.

So if I understand you, you're saying the `Lock` API makes sense to keep, but specifically and only for typed arrays.

On the one hand, it's nice that we can borrow a `JsBox` more ergonomically. On the other, the asymmetry is a little frustrating. It's a shame there's no way (AFAIK) to dynamically borrow an `ArrayBuffer`, which we could then use to make this safe without having to lock the JS VM. You can permanently detach its backing buffer, but that doesn't help this use case; we'd need a temporary, reversible version of detaching.

> The `Lock` statically prevents calling any Neon methods that could potentially mutate the slice.

Exactly! This is definitely subtle (if ingenious—thank Niko for the idea), and is probably worth some better explanation in docs, assuming we keep the API.

If we only end up needing it for `ArrayBuffer` and not `JsBox`, I wonder if we want a less general name than `Borrow`, especially since `JsBox` doesn't use it. Maybe something that makes more explicit that it's a stop-the-world borrow, so the connection to `Lock` is clearer?

> Lastly, there is an edge case where the `Lock` API might not uphold invariants. If the `Buffer` was created as an external buffer (memory created and managed externally), it's possible for other non-Neon native code to mutate these types of buffers. Neon could use n-api methods for detecting externals and refuse to lend them. Additionally, neon could provide `unsafe fn borrow_unchecked` methods for reading data from externals.

Good catch. That makes sense to me.

So, some of this stuff is out of scope for this RFC, but probably worth summarizing the issues in the "open questions" section so we can capture them for subsequent work.

text/0000-jsbox.md
dherman: I feel comfortable making all the APIs return an unwrapped `JsBox<_>`. `NeonResult` is meant to represent whether the operation caused a JS exception. Failing because it was already in a throwing state is basically a programmer error that we happen not to be able to enforce through the type system and therefore a panic; the VM shutting down is extremely rare and unrelated to triggering a JS exception.

text/0000-jsbox.md
kjvalencik: @dherman I like the clear naming of `borrow` for the API. However, we could move  it from `neon::borrow` to `neon::binary` to clarify the relationship to `JsArrayBuffer`.

text/0000-jsbox.md
```diff
@@ -0,0 +1,384 @@
...
+* `napi_get_value_external()`. This API retrieves the external data pointer that was previously passed `to napi_create_external()`.
+* `napi_typeof()`. Returns `napi_external` if the type is an external type.
+
+##### Safety
+
+The `JsBox` API is not entirely safe. It relies on the user only passing N-API externals created within that Neon library. Passing an external created by another library, potentially even another neon library, is _undefined behavior_.
+
+It is *not* undefined behavior to attempt a value that is not an external as a `JsBox`. This will fail predictably. It **is** undefined behavior to downcast a non-`JsBox` externally because the pointer will be treated as if it is one.
+
+Progress is being made to add a [tagging feature](https://github.com/nodejs/node/pull/28237) to more safely unwrap externals. It would allow Neon to tag externals with an id that uniquely identifies that module to safely prevent downcasting if they are unrecognized. It should be incorporated into future designs; however, it is not included in this proposal.
```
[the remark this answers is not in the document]
kjvalencik: I clarified this section a bit further. This is under the "Implementation Notes" section. You are correct that it is an implementation detail.

I am confident that we will be able to make this safe _when the API is available_; however, it has not been backported to all of our supported node versions. We will either need to ship without it or conditionally compile it based on the node version.

Considering that we don't currently have mechanisms in place to conditionally compile based on node version and do not have other use cases in mind for such a feature, we should follow-up later when the API is broadly available. In my opinion, this is a small enough gap that it shouldn't prevent stabilization of `JsBox`.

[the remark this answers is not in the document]
dherman: Awesome, I agree. My only request would be to clarify this point a bit in the RFC text itself. But I'll leave that to you. Other than that one clarification, I'm super happy with this RFC and feel it's ready!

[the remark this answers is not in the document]
dherman: Oh it looks like you clarified, lemme re-read…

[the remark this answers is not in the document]
dherman: Looks great! Thanks so much for the clarifications.

kjvalencik:
This RFC has entered the final comment period.

kjvalencik:
## Future Expansion

### Background

While exploring the `Persistent` RFC, an alternate design that requires manual drops was proposed. Manual drops are very simple for single use `Persistent`, but can become complicated when `Persistent` are cloned.

### Possible Solution

`JsBox` could support an optional `finalizer` that is called on the main JavaScript thread prior to the Rust data being dropped. Since in most cases multi-use `Persistent` will be owned by a `JsBox`, it provides an opportunity to perform a manual drop.

### API

```rust
pub trait Finalize {
    fn finalize(self, cx: FinalizeContext);
}

impl<T: Send + 'static> JsBox<T> {
    pub fn with_finalizer<'a, C>(cx: &mut C, value: T) -> Handle<'a, JsBox<T>>
    where
        C: Context<'a>,
        T: Finalize;
}
```

An additional `JsBox` constructor is added that accepts a `T: Finalize` and calls the `finalize` method prior to dropping.

Following idiomatic patterns establish elsewhere, `Persistent` would implement a `Drop` trait that panics to prevent user error.

```rust
impl<T> Drop for Persistent<T> {
    fn drop(&mut self) {
       panic!("Persistent must be manually dropped.");
    }
}
```

#### Alternative

An alternative design is to provide the `finalize` method as an argument to the constructor instead of as a trait.

```rust
impl<T: Send + 'static> JsBox<T> {
    pub fn with_finalizer<'a, C>(
        cx: &mut C,
        value: T,
        finalizer: fn(FinalizerContext, T),
    ) -> Handle<'a, JsBox<T>>
    where
        C: Context<'a>,
}
```

The advantage of the trait approach is that it co-locates the finalizer with the data. The disadvantage is that Neon would not prevent a user from defining a trait, but failing to use the `with_finalizer` constructor.

In my opinion the `trait` approach is preferred and the ergonomics could be improved in and when the Rust specialization RFC is implemented.

### Example

```rust
struct MyServer {
    server: Server,
    // Wrapping the `Persistent` in a `ManuallyDrop` provides two advantages:
    // 1. Immediately signals that something in this struct needs special attention
    // 2. If a manual `drop` is missed, it causes a leak instead of a `panic`
    callback: ManuallyDrop<Persistent<JsFunction>>,
}

impl Finalize for MyServer {
    fn finalize(self, cx: FinalizeContext) {
        self.callback.drop(&mut cx);
    }
}
```

text/0000-jsbox.md
```diff
@@ -111,45 +111,68 @@ fn get_user(mut cx: FunctionContext) -> JsResult<JsString> {
...
-    let username = pool.get_by_id(id)
+    let username = db.pool().get_by_id(id)
 
     Ok(cx.string(username))
 }
 ```
 
 Data may only be borrowed immutably. However, `RefCell` can be used to introduce interior mutability with dynamic borrow checking rules:
 
+## FIXME: Need to add a blanked impl of `Finalize` on common types like `RefCell`
```
dherman, line 162: We should close this FIXME. If we aren't sure we've thought of all the std types we need blanket impls for, we can leave a note in the open questions section.

dherman (commented):
The changes look good to me! Just one FIXME to update.

kjvalencik:
Merged in https://github.com/neon-bindings/rfcs/commit/951e6257f1766576a5f20f93219ea8c40c9ff931

PR #35, opened 2020-09-17
TaskBuilder and JsPromise

kjvalencik:
API for executing tasks on the `libuv` threadpool and settling JavaScript `Promise`.

```rust
fn return_a_promise(cx: FunctionContext) -> JsResult<JsPromise> {
    let promise = cx
        .task(|| { /* ... */)
        .promise(|mut cx, _result| Ok(cx.undefined()));
    
    Ok(promise)
}
```

Implementation PR: https://github.com/neon-bindings/neon/pull/789

### DIFF
--- a/text/0000-promise.md
+++ b/text/0000-promise.md
@@ -1,0 +1,260 @@
+- Feature Name: `JsPromise`
+- Start Date: 2020-09-17
+- RFC PR: (leave this empty)
+- Neon Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Provide an API for creating, resolving and rejecting JavaScript Promises.
+
+```rust
+fn return_a_promise(cx: FunctionContext) -> JsResult<JsPromise> {
+    let (promise, deferred) = cx.promise();
+    let msg = cx.string("Hello, World!");
+
+    deferred.resolve(&cx, msg);
+
+    Ok(promise)
+}
+```
+
+# Motivation
+[motivation]: #motivation
+
+JavaScript Promise support has been a [long requested feature](https://github.com/neon-bindings/neon/issues/73). Promises are desirable for several reasons:
+
+* Considered best practices in idiomatic JavaScript
+* Enable `await` syntax for asynchronous operations
+* More easily map to asynchronous operations in native code
+
+Additionally, they can be combined with the [`EventQueue`](https://github.com/neon-bindings/rfcs/pull/32) API for very simple asynchronous threaded usage.
+
+```rust
+fn real_threads(cx: FunctionContext) -> JsResult<JsPromise> {
+    let queue = cx.queue();
+    let (promise, deferred) = cx.promise();
+
+    std::thread::spawn(move || {
+        let result = perform_complex_operation();
+    
+        queue.send(move || {
+            let result = cx.number(result);
+
+            deferred.resolve(&cx, result);
+        });
+    });
+
+    Ok(promise)
+}
+```
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Before `Promise`, callbacks of the form `function (err, value)` were very common in JavaScript. Neon has excellent for for these "node style" callbacks in the `Task` trait.
+
+```rust
+fn fibonacci_async(mut cx: FunctionContext) -> JsResult<JsUndefined> {
+    let n = cx.argument::<JsNumber>(0)?.value() as usize;
+    let cb = cx.argument::<JsFunction>(1)?;
+
+    FibonacciTask::new(n).schedule(cb);
+
+    Ok(cx.undefined())
+}
+```
+
+However, in idiomatic JavaScript, this should method should return a promise. This can be solved with some glue code in JavaScript:
+
+```js
+const util = require('util');
+const native = require('../native');
+
+export const fibonacciAsync = util.promisify(native.fibonacci_async);
+```
+
+Alternatively, a `Promise` can be constructed directly in Neon. Unlike an asynchronous method that accepts a callback, a method that returns a promise requires two parts:
+
+* A `Promise` to return from the method
+* A hook to resolve or reject the `Promise`
+
+In JavaScript, this looks like the following:
+
+```js
+function asyncMethod() {
+    return new Promise((resolve, reject) => {
+        if (Math.random() > 0.5) {
+            resolve();
+        } else {
+            reject();
+        }
+    });
+}
+```
+
+In `Neon`, when a `JsPromise` is constructed, a `Deferred` object is also provided. It can be thought of as the following pattern in JavaScript:
+
+```js
+function deferred() {
+    const deferred = {};
+    const promise = new Promise((resolve, reject) => {
+        deferred.resolve = resolve;
+        deferred.reject = reject;
+    });
+
+    return [promise, deferred];
+}
+
+function asyncMethod() {
+    const [promise, d] = deferred();
+
+    setTimeout(() => d.resolve(), 5000);
+
+    return promise;
+}
+```
+
+This could be written in Neon with the following:
+
+```rust
+fn async_method(cx: FunctionContext) -> JsResult<JsPromise> {
+    let queue = cx.queue();
+    let (promise, d) = cx.promise();
+
+    std::thread::spawn(move || {
+        std::thread::sleep(std::time::Duration::from_millis(5000));
+
+        queue.send(|cx| d.resolve(cx.undefined()));
+    });
+
+    Ok(promise)
+}
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+The `JsPromise` API consists of two structs:
+
+* `JsPromise`. Opaque value; only useful for passing back to JavaScript.
+* `Deferred`. Handle for resolving and rejecting the related `Promise`.
+
+They may only be created with the `deferred` method on `Context`.
+
+```rust
+trait Context {
+    fn deferred(&mut self) -> (Handle<JsPromise>, Deferred);
+}
+```
+
+## `JsPromise`
+
+Opaque handle that represents a JavaScript `Promise`. It is not threadsafe.
+
+```rust
+/// A JavaScript `Promise`
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub struct JsPromise(raw::Local);
+
+impl Value for JsPromise {}
+
+impl Managed for JsPromise {
+    fn to_raw(self) -> raw::Local { self.0 }
+    fn from_raw(h: raw::Local) -> Self { JsPromise(h) }
+}
+
+impl ValueInternal for JsPromise {
+    fn name() -> String { "promise".to_string() }
+    fn is_typeof<Other: Value>(env: Env, other: Other) -> bool {
+        unsafe { neon_runtime::tag::is_promise(env.to_raw(), other.to_raw()) }
+    }
+}
+
+unsafe impl This for JsPromise {
+    fn as_this(_env: Env, h: raw::Local) -> Self {
+        JsPromise(h)
+    }
+}
+
+impl JsPromise {
+    pub(crate) fn new_internal<'a>(value: raw::Local) -> Handle<'a, JsPromise> {
+        Handle::new_internal(JsPromise(value))
+    }
+}
+```
+
+## `Deferred`
+
+`Send` handle for resolving or rejecting a `JsPromise`.
+
+```rust
+pub struct Deferred(*mut c_void);
+
+unsafe impl Send for Deferred {}
+
+impl Deferred {
+    /// Resolves a deferred Promise with the value
+    fn resolve<'a, C: Context<'a>, T: Value>(self, cx: &mut C, v: Handle<T>);
+
+    /// Rejects a deferred Promise with the error
+    fn reject<'a, C: Context<'a>, E: Value>(self, cx: &mut C, err: Handle<E>);
+
+    /// Resolves or rejects a deferred Promise with the contents of a `Result`
+    fn complete<'a, C: Context<'a>, T: Value, E: Value>(self, cx: &mut C, result: Result<T, E>);
+}
+```
+
+Similar to [`Persistent`](https://github.com/neon-bindings/rfcs/pull/32), if a `Deferred` is not resolved or rejected, the `Promise` will leak. To help catch this and guide users towards a correct implementation, `Deferred` should `panic` on `Drop` if not used.
+
+```rust
+impl std::ops::Drop for Deferred {
+    fn drop(&mut self) {
+        panic!("JsPromise leaked. Deferred must be used.");
+    }
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None? :grin:
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+## High Level Promise Tasks
+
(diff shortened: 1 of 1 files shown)


text/0000-promise.md
```diff
@@ -0,0 +1,260 @@
...
+    });
+
+    Ok(promise)
+}
+```
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Before `Promise`, callbacks of the form `function (err, value)` were very common in JavaScript. Neon has excellent for for these "node style" callbacks in the `Task` trait.
```
dherman, line 55: s/for for/support for/

dherman: Maybe worth mentioning that the `JsBox` and thread-safe handles RFCs also provide support for asynchronous computations on arbitrary Rust threads, also with classic node-style callbacks.

text/0000-promise.md
```diff
@@ -0,0 +1,260 @@
...
+    let n = cx.argument::<JsNumber>(0)?.value() as usize;
+    let cb = cx.argument::<JsFunction>(1)?;
+
+    FibonacciTask::new(n).schedule(cb);
+
+    Ok(cx.undefined())
+}
+```
+
+However, in idiomatic JavaScript, this should method should return a promise. This can be solved with some glue code in JavaScript:
```
dherman, line 68: s/should method should/method should/

text/0000-promise.md
```diff
@@ -0,0 +1,260 @@
...
+The `JsPromise` API consists of two structs:
+
+* `JsPromise`. Opaque value; only useful for passing back to JavaScript.
+* `Deferred`. Handle for resolving and rejecting the related `Promise`.
+
+They may only be created with the `deferred` method on `Context`.
+
+```rust
+trait Context {
+    fn deferred(&mut self) -> (Handle<JsPromise>, Deferred);
```
dherman, line 147: Elsewhere in the RFC you call this method `promise()`.

dherman: Given the `let (sender, receiver) = channel();` precedent in std, should the order be `let (deferred, promise) = cx.promise();` instead?

text/0000-promise.md
```diff
@@ -0,0 +1,260 @@
...
+}
+
+impl JsPromise {
+    pub(crate) fn new_internal<'a>(value: raw::Local) -> Handle<'a, JsPromise> {
+        Handle::new_internal(JsPromise(value))
+    }
+}
+```
+
+## `Deferred`
```
dherman, line 188: I know the JS world has a historical legacy of "Deferred" being the name for this data structure, but I don't know if that history is being lost over time since the ecosystem now uses standard ES promises, which don't have any concept by this name.

And without the mindshare, I'm not sure the name really conveys much intuitive meaning. Is it worth thinking about alternatives? Maybe something that leans on the "sender" precedent with Rust channels? "Fulfiller"? "Completer"? Those are awkward. I don't have any amazing suggestions right now, but I'll think about it.

text/0000-promise.md
```diff
@@ -0,0 +1,260 @@
...
+
+`Send` handle for resolving or rejecting a `JsPromise`.
+
+```rust
+pub struct Deferred(*mut c_void);
+
+unsafe impl Send for Deferred {}
+
+impl Deferred {
+    /// Resolves a deferred Promise with the value
```
dherman, line 198: Why don't we add:
```rust
/// Resolves a deferred Promise with the value. Equivalent to `self.complete(Ok(v))`.
```

text/0000-promise.md
```diff
@@ -0,0 +1,260 @@
...
+    fn resolve<'a, C: Context<'a>, T: Value>(self, cx: &mut C, v: Handle<T>);
+
+    /// Rejects a deferred Promise with the error
```
dherman, line 201: And here we could add:
```rust
/// Rejects a deferred Promise with the error. Equivalent to `self.complete(Err(err))`.
```

text/0000-promise.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None? :grin:
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
```
dherman, line 226: We should discuss naming alternatives here (e.g. for `Deferred` as well as `cx.deferred()` vs `cx.promise()`) as we come up with them.

text/0000-promise.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+The issue with this approach is that `Deferred` is `Send` and `JsPromise` is *not*. We would also need to provide a getter for the resolve/reject handle. That getter would need to _consume_ self because it cannot be used multiple times. This would result in worse ergonomics.
+
+Instead a tuple is returned, similar to `(sender, receiver)` returned from `std::sync::mpsc::channel`.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should there be a constructor method on `JsPromise`? This is slightly awkward because it's impossible to create a `Promise` without also creating a `Deferred`.
```
dherman, line 260: Yeah it seems OK to me to only have it on the context. We could also have `neon::types::promise(&mut cx)` I suppose, but it's really not providing any value over just `cx.promise()`.

dherman:
This is freaking amazing.

text/0000-promise.md
kjvalencik: That's a great point on the order! I like that. 👍

text/0000-promise.md
kjvalencik: @dherman What do you think about calling it `Sender` to match channels? It's conceptually very similar to a `oneshot`: https://docs.rs/futures/0.3.5/futures/channel/oneshot/index.html.

With that change, instead of separate `resolve` / `reject`, I would opt for a single combined method:

```rust
impl Sender {
    pub fn send<T: Value, E: Value>(result: Result<T, E>) {
        /* ... */
    }
}
```

The downside of this approach is that it requires the user to provide an `Err` type even when the code is infallible, which isn't very ergonomic.

text/0000-promise.md
kjvalencik: Do you have any suggestions?

text/0000-promise.md
kjvalencik: For symmetry, I think there should be a constructor on the type. In this example, `Deferred` is a placeholder for the final name:

```rust
// All equivalent
cx.promise()
JsPromise::deferred(&mut cx)
Deferred::new(&mut cx)
```

kjvalencik: The symmetry should not be included. We can take a look at the `mpsc` API where there is no way to detect a `Sender` or `Receiver`. They are only created with `channel()`.

Similarly, the only way to create `JsPromise` and `Deferred` is with `cx.promise()`.

kjvalencik:
For high level APIs that reduce complexity and closure nesting, I propose:

```
// Wraps with a `try_catch` and resolves/rejects based on the `JsResult`
// Without GAT, because of the lifetime on the result, this might need to return `JsValue` instead of `T: Value`
Deferred::and_settle(self, f: impl FnOnce(cx: TaskContext) -> JsResult);

// On queue, to remove the `.send(move || {})` closure boilerplate:
// Calls `Deferred::and_settle` for you
EventQueue::send(&self, f: Impl FnOnce....)
```

text/0000-promise.md
```diff
@@ -0,0 +1,460 @@
+- Feature Name: `JsPromise` and `TaskBuilder`
+- Start Date: 2020-09-17
+- RFC PR: (leave this empty)
+- Neon Issue: (leave this empty)
+
+## Summary
+[summary]: #summary
+
+Provides an ergonomic API for executing tasks on the libuv threadpool and notifying JavaScript on the main thread with a result.
```
dherman, line 9: This wording caught me off guard at first—I think you're saying the libuv thread pool is specifically the `TaskBuilder` half of the RFC, not the `JsPromise` half, right?

text/0000-promise.md
```diff
@@ -0,0 +1,460 @@
...
+impl std::ops::Drop for Deferred {
+    fn drop(&mut self) {
+        panic!("JsPromise leaked. Deferred must be used.");
+    }
+}
+```
+
+#### `Channel` extension
+
+A new method `Channel::send_and_settle` is added to reduce the boilerplate of nested closures from `Channel::send` and `Deferred::settle_with`. Example without the convenience method:
```
dherman, line 314: I'd propose we call this `Channel::settle_with`, for a few reasons:
- It's a little more ergonomic / less of a mouthful than `send_and_settle`
- It hints at being a shorthand by having the same name as `Deferred::settle_with`
- It's nice for refactoring that you can change from sync to async with a slightly easier-to-understand change

text/0000-promise.md
```diff
@@ -0,0 +1,460 @@
...
+Node-API allows async work to be [cancelled](https://nodejs.org/api/n-api.html#n_api_napi_cancel_async_work) after it has been scheduled, but before it has executed. This is left as future improvement since we do not have a clear use case in mind and it can be added without breaking changes.
+
+Two patterns for adding it without breaking change:
+* Add `fn handle(&self) -> TaskHandle` to `TaskBuilder` to allow creating cancellation handles before scheduling
+* Add methods that return a cancellation handle in addition to the thunk (e.g., `fn cancellable_promise(self, f: F) -> (TaskHandle, Handle<JsPromise>)`
+
+## Unresolved questions
+[unresolved]: #unresolved-questions
+
+* Are the thunk names `TaskBuilder::promise` and `TaskBuilder::complete` clear?
```
dherman, line 457: IMO `TaskBuilder::promise` is very clear.

Re: `complete`, a few possible alternatives:
- `on_complete`
- `and_then` (by analogy to various Rust std APIs)

Not sure what I think, but `complete` doesn't seem bad.

text/0000-promise.md
```diff
@@ -0,0 +1,460 @@
...
+* Is it helpful to mirror deferred constructors across all types instead of only having `cx.promise()`?
```
dherman, line 458: This question slightly bugged me too. On the one hand, I like thinking of all the `cx` constructor methods as convenient shorthands. On the other hand, having `::new()` methods that don't return the type directly but a pair consisting of the type along with a related type feels a little un-idiomatic.

text/0000-promise.md
```diff
@@ -0,0 +1,460 @@
...
+* Is `cx.promise()` a good name or would `cx.deferred()` be better?
```
dherman, line 459: FWIW I think `promise()` is the main thing you're trying to construct, so that's my preference.

dherman (changes requested):
A couple small suggestions, but it's very close!

text/0000-promise.md
kjvalencik: That's correct. I think it will be more clear if I break it into two sentences since the next line explains why they are being designed together.

text/0000-promise.md
kjvalencik: I like `and_then` since it matches common naming in Rust, for example, in the [`futures` crate](https://docs.rs/futures/0.3.17/futures/future/trait.TryFutureExt.html#method.and_then).

text/0000-promise.md
kjvalencik: I think it's best to remove `JsPromise::new` and `Deferred::new` and only have `cx.promise()`. This matches the design of mpsc:

* in [`std::sync::mpsc`](https://doc.rust-lang.org/std/sync/mpsc/index.html)
* and [`crossbeam`](https://docs.rs/crossbeam/0.8.1/crossbeam/channel/index.html)

We can document `cx.promise()` in the struct docs for `JsPromise` and `Deferred`.

text/0000-promise.md
kjvalencik: I agree. I think it makes it more discoverable and the borrow checker will let the user know they getting something other than just a `JsPromise`.

dherman (approved):
🎉

kjvalencik:
A couple of potential issues with this RFC.

## Lifetime Inference

Since the closure in `settle_with` takes a borrowed `TaskContext`, users may run into a lifetime inference issue if the closure is saved to a variable instead of passed directly to the function.

```
395 |                 .try_settle_with(deferred, callback)
    |                  ^^^^^^^^^^^^^^^ one type is more general than the other
    |
    = note: expected enum `Result<neon::handle::Handle<'_, _>, _>`
               found enum `Result<neon::handle::Handle<'a, _>, _>`
```

This issue is avoided if an owned `TaskContext` is passed. Since Neon owns the lifetime, we can give an owned `TaskContext` to ease this case. We may also want to consider doing the same in `try_catch`.

This is a breaking change and needs to be decided prior to merging.

## Promises and synchronous exceptions

One advantage of using `async function() {}` over a `function() {}` that returns a promise is that the function is guaranteed to *always* return a `Promise`. Synchronous exceptions are transparently converted into rejections. This is cumbersome to do in Neon because it requires using `try_catch` and handling the result. I propose a new helper `JsPromise::try_catch` for performing this.

```rust
fn never_throws_synchronously(mut cx: FunctionContext) {
    Ok(JsPromise::try_catch(|cx| {
        let n = cx.argument::<JsNumber>(0)?.value(cx);
        
        cx.task(move || n + 42)
            .promise(|cx, n| Ok(cx.number(n)))
    }))
}
```

The downside of this approach is that it still requires careful code to avoid an `unhandledRejection` when using `cx.promise()`. The `(deferred, promise)` pair shouldn't be created until after all code that can be thrown synchronously.

Alternatively, we could provide a `Deferred::try_catch` that uses the existing `promise` to avoid this gotcha. However, I'm not sure how to align these two use cases. A user could still stumble onto the `JsPromise::try_catch` version which is better suited towards `TaskBuilder::promise`.

```rust
fn never_throws_synchronously(mut cx: FunctionContext) {
    Ok(cx.try_promise(|cx, deferred| {
        let n = cx.argument::<JsNumber>(0)?.value(cx);
        let channel = cx.channel();
        
        std::thread::spawn(move || {
            let n = n + 42;
            
            channel.settle_with(deferred, move |cx| Ok(cx.number(n)));
        });

        Ok(())
    }))
}
```

This is an extension and can be added later without causing any breaking changes.

dherman:
@kjvalencik So you mean `settle_with` and `try_settle_with` should take an owned `TaskContext`, yes? That seems fine to me.

> We may also want to consider doing the same in `try_catch`.

Are you talking about a backwards-incompatible change to the existing `cx.try_catch()` API, or are you talking about one of the `try_catch` proposals you're making in the second half of your comment? I'm not quite following the "Promises and synchronous exceptions" section, it seems like there are several different hypothetical APIs in play here, some of which might be typos but I'm not sure?

- `JsPromise::try_catch`
- `Deferred::try_catch`
- `Context::try_promise`

This sentence was hard to follow in particular:

> A user could still stumble onto the `JsPromise::try_catch` version which is better suited towards `TaskBuilder::promise`.

kjvalencik:
@dherman I meant a breaking change to `cx.try_catch` (which is still feature flagged).

These are several different proposals for similar things. The idea is we should make it easy for a user to write a Neon function that always resolved or rejects a promise instead of throwing an error synchronously.

The issue I'm having is I haven't been able to come up with a single API that works for both `TaskBuilder` and `Context::deferred`.

kjvalencik:
After some testing, moving to an owned `TaskContext` does not resolve this issue. With that said, I think it should still get an owned `TaskContext` for consistency.

kjvalencik:
Merged in https://github.com/neon-bindings/rfcs/commit/2e6c1f1f9a594a636a8206665d6263652284ed1b

PR #36, opened 2020-09-17
RFC: Streamlined UX

dherman:
This RFC proposes streamlining the directory layout and command-line UX of Neon projects, making the design simpler, more intuitive, and easier to learn and adopt.

In particular, starting a new Neon project looks like:

```
$ npm init neon my-project
...
✨ Created Neon project my-project. Happy 🦀 hacking! ✨
$ cd my-project
$ npm install
$ node -e 'console.log(require(".").hello())'
hello node
```

And the directory layout is flattened and simplified to:

```
my-project/
├── .gitignore
├── Cargo.toml
├── README.md
├── package.json
└── src
    └── lib.rs
```

[Rendered RFC](https://github.com/neon-bindings/rfcs/blob/70edc0d168e1cbc175574b427af6674d73f7d377/text/0000-streamlined-ux.md)


text/0000-native-cli.md
```diff
@@ -0,0 +1,136 @@
...
+- `npm init neon-electron-app <args...>`
+
+respectively.
+
+## Replacing `neon build`
+
+There are a couple of key developments that will allow us to move past `neon build`:
+
+1. Once we move to the [N-API backend](https://github.com/neon-bindings/neon/issues/444), we no longer need to invalidate a Neon build when changing between versions of Node. This means we won't need to use the custom invalidation logic (which has proved to be fragile and a source of user frustration).
+2. We discovered that we can skip the final step of copying the built DLL into a `.node` file by replacing the JS boilerplate that loads the addons with a call to the lower-level but official [`process.dlopen()`](https://nodejs.org/api/process.html#process_process_dlopen_module_filename_flags) API.
```
kjvalencik, line 91: Unfortunately, `process.dlopen()` may not be usable. Cargo build scripts are intentionally kept from knowing the location of the `target` directory. This makes it difficult for our generated `index.js` to find the lib file.

Instead, we can pass a `rustc` flag to output the library in the correct location. This is possible because build scripts _can_ know the location of the project directory.

```rust
// in build.rs
let manifest_dir = env::var_os("CARGO_MANIFEST_DIR").unwrap();
let output = Path::new(&manifest_dir).join("index.node");

println!("cargo:rustc-cdylib-link-arg=-o");
println!("cargo:rustc-cdylib-link-arg={}", output.display());
```

This sample code has been tested and works, but would need to be improved with platform dependent code to name the file correctly in Windows.

text/0000-native-cli.md
```diff
@@ -0,0 +1,136 @@
...
+
+To create a new Neon library, use the command:
+
+```
+npm init neon-lib my-lib
+```
+
+Use whatever name you want for your library in place of `my-lib`.
+
+### Node apps
```
kjvalencik, line 41: Can you elaborate on how an app structure might differ from a library?

My recommendation is that we strongly encourage a monorepo pattern with `neon-app`:

```
my-app/
├── Cargo.toml
├── neon-module
│   ├── Cargo.toml
│   ├── index.js
│   ├── package.json
│   └── src
│       └── lib.rs
├── package.json
└── src
    └── index.js
```

The top level project is the main app. The neon-module is a flat and thin.

text/0000-native-cli.md
```diff
@@ -0,0 +1,136 @@
...
+
+To create a new Node app, use the command:
+
+```
+npm init neon-app my-app
+```
+
+Use whatever name you want for your library in place of `my-app`.
+
+### Electron apps
```
kjvalencik, line 51: It might be interesting to explain why electron apps are treated differently from normal apps.

It would be possible to move `electron-build-env` functionality into `build.rs` and automate the discovery of an `electron` dependency. I would not recommend this because of the complexity and potential for false positives, but it might be useful to explain why we did not choose this.

It's also possible that N-API combined with dynamic library loading, we might not need any special code to handle Electron.

text/0000-native-cli.md
```diff
@@ -0,0 +1,136 @@
...
+
+1. Once we move to the [N-API backend](https://github.com/neon-bindings/neon/issues/444), we no longer need to invalidate a Neon build when changing between versions of Node. This means we won't need to use the custom invalidation logic (which has proved to be fragile and a source of user frustration).
+2. We discovered that we can skip the final step of copying the built DLL into a `.node` file by replacing the JS boilerplate that loads the addons with a call to the lower-level but official [`process.dlopen()`](https://nodejs.org/api/process.html#process_process_dlopen_module_filename_flags) API.
+
+In order to make `cargo build` work from the root directory of a Neon project, we can put a `Cargo.toml` in the root directory that either defines a [Cargo workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) or that shells out to running `cargo build` in the proper subdirectory. Alternatively, we might just try experimenting with flattening the directory structure so that the Cargo directory _is_ the root directory. We should experiment with prototypes to see what looks and feels best for the developer experience.
```
kjvalencik, line 93: My recommendation is to flatten the directory.

Placing a workspace `Cargo.toml` in the root adds an extra step for users with a monorepo that would like the neon module to be part of that workspace. First, they need to delete the `Cargo.toml` at the project root.

I'm not clearing how a build that shells out would work.

text/0000-native-cli.md
```diff
@@ -0,0 +1,136 @@
...
+(Note that as a _programming_ syntax, this might arguably be preferable for the visual distinction. But the command-line is meant to be quick and easy to type and remember.) Another option would be to support both as aliases for each other, but this might just overwhelm users with pointless choices.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- What other build invalidation logic, if any, do we still need to add? Can this all be done within `build.rs`? We should review the `neon build` invalidation logic and check for anything this RFC has overlooked.
+- What is the best way for `cargo build` to work in the root?
+  * `cargo build --all`?
+  * `cargo build` by shelling out to `cargo build` in the right subdirectory?
+  * Flatten the directory structure so the root directory is both a Node package and a Rust crate? (This is an attractive option, as it simplifies the directory structure e.g. in the tree view of an IDE, and generally may make the cognitive load of a Neon project feel lighter.)
```
kjvalencik, line 124: 👍  I like this the best. More complex project layouts can use a mono-repo to be more expressive. For example, instead of TypeScript and Rust needing to content for the coveted `src` directory, they end up moved into separate modules.

text/0000-native-cli.md
```diff
@@ -0,0 +1,136 @@
...
+- On the topic of lightening the directory layout, should we also put the `index.js` into the root directory? This might look like:
```
kjvalencik, line 125: We might not need `index.js` at all in a flat structure since the generated `index.node` would be in this location.

kjvalencik (commented):
Something we may want to consider adding to this scope is making the `fn build` function for N-API part of the `neon` module.

It has zero dependencies and would allow users to use neon with _only_ `neon` in their `Cargo.toml`.

I'm not sure how to accomplish it, but it would also be neat if building `neon` would fail with a helpful message if the user was missing a `build.rs`. This could allow for a workflow where users can `cargo add neon` and "follow the compiler."

text/0000-native-cli.md
```diff
@@ -0,0 +1,136 @@
...
+
+Neon provides templates for conveniently creating various types of Neon projects, including:
+
+- libraries;
+- Node apps; and
+- Electron apps.
+
+### Libraries
+
+To create a new Neon library, use the command:
```
kjvalencik, line 33: # Structure

```
.
├── Cargo.toml
├── README.md
├── build.rs
├── index.node
├── package.json
├── src
│   └── lib.rs
```

text/0000-native-cli.md
dherman: Can you help me understand this part? I don't quite understand how dynamic library loading could accomplish that. And if we can't, it seems like we don't yet fully have a story for how to deprecate `neon build`?

text/0000-native-cli.md
dherman: Re: shelling out, I meant a `build.rs` in the root directory that shells out to `cargo build` from the subdirectory. But this is probably a really bad idea. :) I'll just delete the idea from the RFC.

text/0000-native-cli.md
dherman: Removed in latest draft!

text/0000-native-cli.md
dherman: Latest draft proposes and describes the flattened structure.

text/0000-native-cli.md
dherman: Incorporated in the latest draft.

kjvalencik (approved):

text/0000-native-cli.md
```diff
@@ -0,0 +1,103 @@
...
+- **Simplified layout:** A Neon project no longer contains a mix of JS and Rust subdirectory structures; it's flattened into a standard Cargo project + a `package.json` to treat it as a Node native module.
+
+This design lightens the cognitive burden of learning Neon, making it less intimidating for new users and easier to use for everyone.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## Creating a new Neon project
+
+The `npm init neon` command makes it easy to start a new Neon project. You can use `npm init neon my-app` to create a new `my-app` Neon directory, or `npm init neon` to initialize an existing directory as a Neon project.
```
kjvalencik, line 27: Nit, we _just_ discussed potentially eliminating appending `my-app` on the initial revision. We may want to consider changing this to further expansion.

text/0000-native-cli.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Neon project layout
+
+A Neon project looks like a typical Rust project that you might create with `cargo new`, but it also contains a `package.json` manifest with the proper npm scripts to build the project as a Node module with `npm install`. After building, a Neon project directory looks like this:
+
+```
+├── Cargo.toml
+├── README.md
+├── build.rs
```
kjvalencik, line 36: Please remove this. There is no longer a `build.rs`

text/0000-native-cli.md
```diff
@@ -0,0 +1,103 @@
...
+├── index.node
+├── package.json
+└── src
+    └── lib.rs
```
kjvalencik, line 40: Nit, this is missing dot files. I think `.gitignore` is the only one we need by default. Supporting ignore files for other version control systems can be left for follow-ups.

kjvalencik (changes requested):

text/0000-native-cli.md
dherman: Agreed, and from testing it out just now (at least with npm 6), it looks like standard `npm init` doesn't create either `.gitignore` or `.npmignore`.

dherman:
@kjvalencik Thanks for the review, I addressed all your feedback.

kjvalencik (approved):

kjvalencik (approved):

PR #38, opened 2020-12-09
Symbol primitive

chrisbajorin:


text/0000-symbol-primitive.md
```diff
@@ -0,0 +1,120 @@
...
+NOTE: Node 10.23 [does not support](https://node.green/#ES2019-misc-Symbol-prototype-description) `Symbol.prototype.description`, so I believe this would need to be feature-flagged.
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+The proposed `new` function has the caveat that you need to add a type annotation to the `None` variant. I'm unsure of how commonplace it is to construct symbols without descriptions. It might be worth considering two constructors, allowing users to skip wrapping the description in `Option`.
+```rust
+impl JsSymbol {
+    pub fn new<'a, C: Context<'a>, S: AsRef<str>>(cx: &mut C, s: S) -> Handle<'a, JsSymbol>;
+    pub fn new_without_description<'a, C: Context<'a>>(cx: &mut C) -> Handle<'a, JsSymbol>;
```
kjvalencik, line 104: I prefer this API to the version that uses `Option`. It seems reasonable to encourage the use of a description and optimize for the common case.

One small suggestion is to rename these `with_description` and `new` respectively.

```rust
impl JsSymbol {
    pub fn new<'a, C: Context<'a>>(cx: &mut C) -> Handle<'a, JsSymbol>;
    pub fn with_description<'a, C: Context<'a>, S: AsRef<str>>(cx: &mut C, s: S) -> Handle<'a, JsSymbol>;
}
```

text/0000-symbol-primitive.md
```diff
@@ -0,0 +1,120 @@
...
+let symbol_without_description = JsSymbol::new(&mut cx, None::<&str>);
+let description = symbol_without_description.description(&mut cx);
+assert_eq!(description, None);
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+#### JsSymbol
+- `new` would be implemented using [napi_create_symbol](https://nodejs.org/docs/latest-v14.x/api/n-api.html#n_api_napi_create_symbolhttps://nodejs.org/docs/latest-v14.x/api/n-api.html#n_api_napi_create_symbol) (as well as the v8 equivalent for the legacy-api)
```
kjvalencik, line 55: I think implementing `Symbol` for the legacy runtime should be considered _optional_ and left to the implementers discretion.

text/0000-symbol-primitive.md
```diff
@@ -0,0 +1,120 @@
...
+        JsSymbol::new(self, s)
+    }
+}
+```
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+This RFC does not currently propose anything for the `Symbol` object wrapper API, which includes Well-Known Symbols and the Symbol global registry (`Symbol.for()`/`Symbol.keyFor()`).
```
kjvalencik, line 119: I completely agree! The first version is best left to features that can be implemented with thin abstractions on N-API.

text/0000-symbol-primitive.md
```diff
@@ -0,0 +1,120 @@
...
+```rust
+impl JsSymbol {
+    pub fn new<'a, C: Context<'a>, S: AsRef<str>>(cx: &mut C, s: S) -> Handle<'a, JsSymbol>;
+    pub fn new_without_description<'a, C: Context<'a>>(cx: &mut C) -> Handle<'a, JsSymbol>;
+    pub fn description<'a, C: Context<'a>>(self, cx: &mut C) -> Option<String>;
+}
+
+trait Context<'a>: ContextInternal<'a> {
+    fn symbol<S: AsRef<str>>(&mut self, s: S) -> Handle<'a, JsSymbol> {
+        JsSymbol::new(self, s)
```
kjvalencik, line 110: The `cx.symbol(...)` version should be common case that includes a description. If renamed above, this would be:

```rust
JsSymbol::with_description(self, s)
```

kjvalencik (commented):
@chrisbajorin This is fantastic! Thanks so much for the contribution.

One recommended addition to this RFC, which is related in usefulness, but not directly tied to it, is adding a `Eq` implementation for all `Value` that leverages `napi_strict_equals`.

chrisbajorin:
@kjvalencik  In regards to `Eq`, are you just looking for a mention that strict equality should hold for all primitive values, or is this something that fits within the reference-level explanation? I see `Eq` currently has a blanket implementation on `Handle<T>`, so I'm not quite sure how this fits in code-wise.

chrisbajorin:
Updated to reflect our discussion in https://github.com/neon-bindings/neon/pull/761

kjvalencik:
@chrisbajorin This looks really great! I'm going to put it into final comment period! Thanks for your contributions!

dherman:
BTW, I just wanted to note that the `description()` method can be technically thought of as a convenience method, since you can implement it with other primitives, i.e. the Neon equivalent of:

```js
const getter = Object.getOwnPropertyDescriptor(Symbol.prototype, 'description').get;
getter.call(sym)
```

Not a crucial point but it makes me feel confident that the `description()` method is a nice convenience to have and doesn't fundamentally add any complexity.

PR #39, opened 2021-03-10
RFC: `ResultExt` trait

davidbailey00:
This follows discussion in https://github.com/neon-bindings/examples/pull/67

I'm quite new to Rust so hopefully my proposal is clear and I haven't missed any obvious alternatives!

[View Rendered RFC](https://github.com/davidbailey00/neon-rfcs/blob/result-ext-trait/text/0000-result-ext-trait.md)

### DIFF
--- a/text/0000-result-ext-trait.md
+++ b/text/0000-result-ext-trait.md
@@ -125,7 +125,9 @@ In the second example under [Guide-level explanation], we create a new `Result` 
 
 > Why is this design the best in the space of possible designs?
 
-I believe this is the most elegant way to 'translate' Rust errors to JavaScript errors. As I understand, it is not possible to automatically convert these errors because the context object is needed to create a JavaScript error, although this would be ideal if possible.
+I believe this is the most elegant way to 'translate' Rust errors to JavaScript errors as it does not require repetitive code.
+
+As I understand, it is not possible to automatically convert these errors because the context object is needed to create a JavaScript error, although this would be ideal if possible.
 
 > What other designs have been considered and what is the rationale for not choosing them?
 
--- a/text/0000-result-ext-trait.md
+++ b/text/0000-result-ext-trait.md
@@ -7,7 +7,7 @@
 
 [summary]: #summary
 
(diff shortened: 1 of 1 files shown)


kjvalencik:
@davidbailey00 The main blocker for this RFC is that in order to support this in a very general fashion, requires Generic Associated Types or GATs, which do not exist in Rust.

We would need to choose between supporting generic `Into` responses on structs or boxed trait object representations (runtime cost).

This is an interesting idea, but I think it needs a proof of concept that can demonstrate we can provide this in a zero-cost, ergonomic format. I then think we would want to do a holistic update of Neon APIs to expect `Into<Throw>` or `Into<Value>` like traits.

davidbailey00:
> The main blocker for this RFC is that in order to support this in a very general fashion, requires Generic Associated Types or GATs, which do not exist in Rust.

On that basis, should this be revisited when https://github.com/rust-lang/rust/issues/44265 lands?

kjvalencik:
@davidbailey00 I would ❤️  to implement this sooner, if possible. I tried a few different hacks to work around the problem, but was unsuccessful. It would be awesome of some contributor could figure out how to solve it.

Maybe there's something we could do with proc macros to make it ergonomic without GATs? Here's my ideal API for `JsFunction::new`:

```rust
trait TryIntoJs {
    type Error: IntoThrow;

    fn try_into_js<'a, C: Context<'a>>(self, cx: &mut C) -> Result<Handle<'a, Value>, Self::Error>;
}

trait IntoThrow {
    fn try_into_js<'a, C: Context<'a>>(self, cx: &mut C) -> JsResult<Self>;
}

impl JsFunction {
    pub fn new<'a, C, F, T, E>(f: F) -> JsResult<'a, JsFunction>
    where
        C: Context<'a>,
        F: Fn(FunctionContext) -> Result<T, E>,
        T: TryIntoJs,
        E: IntoThrow,
    {
        todo!()
    }
}
```

kjvalencik:
Thanks for the contribution! This idea has been incorporated into https://github.com/neon-bindings/rfcs/pull/44.

PR #43, opened 2021-06-03
RFC: add pyo3-like attribute macros to define nodejs classes and methods

Mossaka:
Signed-off-by: Joe Zhou <jiazho@microsoft.com>

### DIFF
--- a/text/0000-adding-pyo3-like-attribute-macros.md
+++ b/text/0000-adding-pyo3-like-attribute-macros.md
@@ -1,0 +1,75 @@
+- Feature Name: adding_pyo3_like_attribute_macros_for_classes_and_methods
+- Start Date: 2021-06-02
+- RFC PR: (leave this empty)
+- Neon Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC proposes adding [PyO3-like](https://pyo3.rs/v0.13.2/class.html) attribute-macros to define JavaScript classes as Rust structs. 
+
+# Motivation
+[motivation]: #motivation
+
+At the moment of migration to `napi`, there is no idiomatic way of writing JavaScript classes in pure Rust. This proposal adds the attribute macros to help users to write classes and methods easily. 
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+I will describe some potential use cases here:
+```rust
+#[neon::class]
+struct Person {
+    name: string,
+    age: u8,
+}
+#[neon::methods]
+impl Person {
+    fn greet(&self, cx: FunctionContext) -> JsResult<JsString> {
+        Ok(cx.string(format!("Hello, my name is {}", self.name)))
+    }
+
+    #[neon::constructor]
+    fn new(mut cx: FunctionContext) -> Self {
+        let name = cx.argument::<JsString>(0)?.value(&mut cx);
+        let age = cx.argument::<JsNumber>(1)?.value(&mut cx);
+        Person { name, age }
+    }
+}
+```
+
+Then we can use `Person` class as the following
+```JavaScript
+let p = Person("Peter", 25);
+p.greet(); // Hello, my name is Peter
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+This is the technical portion of the RFC. Explain the design in sufficient detail that:
+
+- Its interaction with other features is clear.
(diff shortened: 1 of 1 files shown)


text/0000-adding-pyo3-like-attribute-macros.md
```diff
@@ -0,0 +1,75 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+I will describe some potential use cases here:
+```rust
+#[neon::class]
+struct Person {
+    name: string,
+    age: u8,
+}
+#[neon::methods]
```
kjvalencik, line 26: I think these should be namespaced to help prevent conflicts with other macros. This could be named `neon::class_methods` or we could have everything in a `neon::class` module.

```rust
#[neon::class::data]
#[neon::class::methods]
```

text/0000-adding-pyo3-like-attribute-macros.md
```diff
@@ -0,0 +1,75 @@
...
+impl Person {
+    fn greet(&self, cx: FunctionContext) -> JsResult<JsString> {
+        Ok(cx.string(format!("Hello, my name is {}", self.name)))
+    }
+
+    #[neon::constructor]
```
kjvalencik, line 32: The current `declare_types` macro has a two-part initialization with the second part being optional.

1. (Required). Constructor. This creates the Rust struct that gets wrapped into the JS class.
2. (Optional). Initialization. This has access to the _created_ class and can perform actions like assigning properties. This can be thought of as using `this` in a JS constructor.

I think this is valuable and it would be nice to maintain that functionality. I can think of a couple ways to implement it:

1. Another macro attribute like `#[neon::class_init]` for that phase
2. Need to call a function and return a class instead of the data type.

```rust
create_class(Person { name, age })
```

text/0000-adding-pyo3-like-attribute-macros.md
```diff
@@ -0,0 +1,75 @@
...
+Why should we _not_ do this?
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+- Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this?
+
+# Unresolved questions
```
kjvalencik, line 70: Some unresolved questions that I have.

1. What is the name of the type? Do we generate complete new types or do we have something like `JsClass<Person>`?
2. Do we implicitly create a `RefCell` or do we only allow `&self`? I think we should start simple and class methods can *only* take `&self`. If they take a different form of `&self` we fail to compile and if they don't take `self` at all we create a static method.
3. Are there any other _critical_ features? In my opinion we should start simple and not try to add everything.
4. What happens if the class is called as a function instead of a constructor?

kjvalencik (commented):
Thanks so much for this proposal! `Py03` is a great example because a lot of the problem space are shared.

text/0000-adding-pyo3-like-attribute-macros.md
Mossaka: Currently neon only has one proc/attribute macro, right? `main`

Mossaka:
Updated RFC with the comments.

text/0000-adding-pyo3-like-attribute-macros.md
kjvalencik: That's correct.

Mossaka: I actually prefer the simplicity of `#[neon::methods], #[neon::classes]` if in a foreseeable future there won't be any namespace conflicts.

louneskmt:
Any update on this RFC?

kjvalencik:
@louneskmt No update. This RFC itself is blocked by being able to create classes (i.e., a type safe wrappers around [`napi_define_class`](https://nodejs.org/api/n-api.html#napi_define_class) and [`napi_wrap`](https://nodejs.org/api/n-api.html#napi_wrap)).

While this feature is highly desirable, it's not considered a blocker for Neon 1.0 because the functionality can be accomplished with some JavaScript glue code. Blockers for 1.0 have taken priority.

Contributions would be welcome. I have spent a bit of time thinking about what Neon's `napi_define_class` might look like. Most likely:
* A `trait` providing a `constructor`, `name`, and properties
* An instance data cache that can lookup constructors based on the implementing type
* Some code for wrap/unwrap that operates similar to `JsBox`

PR #44, opened 2021-07-20
Borrow API v1.0 Proposal

kjvalencik:
Neon should provide APIs for _safely_ and _ergonomically_ borrowing binary data. Borrowing should leverage the type system for statically checking safety as much as possible without sacrificing flexibility.

[Rendered](https://github.com/neon-bindings/rfcs/blob/kv/borrow/text/0000-node-api-borrow.md)

Implementation: https://github.com/neon-bindings/neon/pull/780
Pre-release published in [`0.10.0-alpha.1`](https://github.com/neon-bindings/neon/releases/tag/0.10.0-alpha.1).

### DIFF
--- a/text/0000-node-api-borrow.md
+++ b/text/0000-node-api-borrow.md
@@ -68,7 +68,7 @@ pub trait TypedArray: private::Sealed {
     fn try_borrow<'a: 'b, 'b, C>(
         &self,
         lock: &'b Lock<'b, C>,
-    ) -> Result<Ref<'b, Self::Item>, TypedArrayError>
+    ) -> Result<Ref<'b, Self::Item>, BorrowError>
         where
             C: Context<'a>;
 
@@ -81,7 +81,7 @@ pub trait TypedArray: private::Sealed {
     fn try_borrow_mut<'a: 'b, 'b, C>(
         &mut self,
         lock: &'b Lock<'b, C>,
-    ) -> Result<RefMut<'b, Self::Item>, TypedArrayError>
+    ) -> Result<RefMut<'b, Self::Item>, BorrowError>
         where
             C: Context<'a>;
 }
@@ -208,11 +208,11 @@ pub struct RefMut<'a, T> {
     ledger: &'a RefCell<Ledger>,
 }
 
-pub struct TypedArrayError {
+pub struct BorrowError {
     _private: (),
 }
 
-impl Error for TypedArrayError {}
+impl Error for BorrowError {}
 ```
 
 The `Ref` and `RefMut` borrow guards serve two purposes:
@@ -230,7 +230,7 @@ impl TypedArray for JsArrayBuffer {
 }
 ```
 
-The `TypedArrayError` type may also be converted to an exception using the *new* `neon::result::ResultExt` trait:
+The `BorrowError` type may also be converted to an exception using the *new* `neon::result::ResultExt` trait:
 
 ```rust
 /// Extension trait for converting Rust [`Result`](std::result::Result) values
@@ -296,7 +296,7 @@ neon
     ├── JsTypedArray<T>
     └── buffer
         ├── TypedArray
-        ├── TypedArrayError
+        ├── BorrowError
         ├── Ref
         └── RefMut
 ```
--- a/text/0000-node-api-borrow.md
+++ b/text/0000-node-api-borrow.md
@@ -37,16 +37,16 @@ There are a few methods on the `Context` trait that do not take `&mut self` and 
 
 The borrowing design consists of a statically checked, ergonomic API and a more complex, but powerful runtime checked API.
 
-#### `trait Borrow`
+#### `trait TypedArray`
 
 ```rust
-pub trait Borrow: private::Sealed {
+pub trait TypedArray: private::Sealed {
     type Item;
 
     /// Statically checked immutable borrow of binary data.
     ///
     /// This may not be used if a mutable borrow is in scope. For the dynamically
-    /// checked variant see [`Borrow::try_borrow`].
+    /// checked variant see [`TypedArray::try_borrow`].
     fn as_slice<'a: 'b, 'b, C>(&'b self, cx: &'b C) -> &'b [Self::Item]
         where
             C: Context<'a>;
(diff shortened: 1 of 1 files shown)


dherman:
Does Node-API have an API for detaching the backing buffer out of an `ArrayBuffer`? That could be another future extension.

dherman:
Seems like another drawback is you can’t mutable borrow multiple buffers (with eg a dynamic check they don’t alias or overlap). I don’t know if this is a rare enough use case not to matter.

dherman:
@kjvalencik Remind me, does changing the scoped methods to `&mut` still allow multiple levels of nested scoped computations? I think there’s no reason that shouldn’t work, right?

kjvalencik:
> Does Node-API have an API for detaching the backing buffer out of an `ArrayBuffer`? That could be another future extension.

Yes, it does with the following constraints:

* It must not have already been detached
* It must be externally allocated

Unfortunately, this wouldn't work with our current API design. We accept anything that is `AsMut<[u8]>` when creating external `ArrayBuffer`. Since this could be _anything_, it's important that we run the `Drop`. However, it's not possible to get the `data` pointer back _out_ of an external buffer. We would need to have that feature added to Node-API.

> Seems like another drawback is you can’t mutably borrow multiple buffers (with eg a dynamic check they don’t alias or overlap). I don’t know if this is a rare enough use case not to matter.

It's a drawback of the "simple API" and the reason that the proposal includes a dynamically checked part. It's covered in this section: https://github.com/neon-bindings/rfcs/pull/44/files#diff-f7efd300745a6d0dcd2273fa31a292295151b581fb63b5ae24ef8661f1924520R71

> @kjvalencik Remind me, does changing the scoped methods to &mut still allow multiple levels of nested scoped computations? I think there’s no reason that shouldn’t work, right?

Yes, it does because the scope creates a new `Context` and then you can create another nested scope off of that.

text/0000-node-api-borrow.md
```diff
@@ -0,0 +1,459 @@
...
+* `Context::compute_scoped`
+
+### API
+
+The borrowing design consists of a statically checked, ergonomic API and a more complex, but powerful runtime checked API.
+
+#### `trait Borrow`
+
+```rust
+pub trait Borrow: private::Sealed {
```
dherman, line 43: Maybe somewhere this could link to the Rust style guide to explain the "sealed" pattern.

text/0000-node-api-borrow.md
```diff
@@ -0,0 +1,459 @@
...
+impl<T: Copy> Borrow for JsTypedArray<T> {
+    type Item = T;
+
+    /* ... */
+}
+```
+
+However, since `JsTypedArray` may overlap with each other, it is important that `Lock` properly track the _region_ of memory and not only the starting pointer. `Lock` will maintain a set of pointer `Range` to track which buffers are currently borrowed.
+
+### Module Layout
```
dherman, line 311: We had a great conversation today about the naming and layout. Here were the main design constraints we identified:

- Commonly used JS types should be in `neon::types` so there's one main place to look in the docs to understand how to use standard JS types.
- Rarely-used datatypes that are specific to typed arrays should not pollute `neon::types` so that the experience of reading the docs for standard JS types isn't overwhelming and full of rarely-needed domain-specific details.
- Anything that users are likely to need to bring into scope for common use cases should not have a name that conflicts with a common Rust standard library.
- Concepts analogous to the Rust standard library should nevertheless use naming conventions that are suggestive of those analogies, to help learnability.
- Neon modules should have names that match normal intuitions as users are likely to think about them, not abstract or theoretical concepts.
- Types that are part of generic Neon functionality (e.g. contexts) should not have functionality specific to typed arrays.
- "Punning" inherent methods after trait methods is useful for making common, idiomatic patterns more ergonomic; the extra boilerplate in the implementation is worth it for the win to user experience.
- Abstractions relating to typed arrays should not have overly generic names.
- The commonly-used name for the typed arrays space is "typed arrays" :)
- The common intuition for the kind of data structure in this space is the "buffer" (`Buffer`, `ArrayBuffer`, `SharedArrayBuffer`)
- Module names that are a noun should be singular since this seems to be idiomatic in the Rust ecosystem (`types` was a special case since `type` is a reserved word)

With that, the ideas we came up with for the naming and layout were:

- Move `Lock` into `neon::context` since it's a generic concept relating to the state of the VM
  - Remove the inherent methods relating to typed arrays, especially since we want to encourage `buffer.borrow()` instead of `lock.borrow()` as the idiomatic pattern.
- Rename `Borrow` to `TypedArray` and `BorrowError` to `TypedArrayError`
- Add inherent methods to the typed array types so users don't need to import `TypedArray` unless they're writing abstractions that need a generic trait
- Keep `JsArrayBuffer`, `JsBuffer`, and `JsTypedArray` in `neon::types`
- Keep `Ref` and `RefMut` named as is and in `neon::types::binary`
- Rename `neon::types::binary` to `neon::types::buffer`

Separately, we also discussed the idea of moving all error types into `neon::result` and renaming that module to `neon::error`, but that's out of scope for this RFC. Since it's a backwards-incompatible change, we should tackle that soon and avoid releasing this (also backwards-incompatible) RFC until we release that one as well, to avoid the number of incompatible upgrades users need to go through. But still it's good to land the PRs individually so we'll separate that into a followup.

dherman (changes requested):
This is a great RFC. I left a detailed comment with the results of the conversation we had today about naming and API organization.

text/0000-node-api-borrow.md
kjvalencik: > Add inherent methods to the typed array types so users don't need to import `TypedArray` unless they're writing abstractions that need a generic trait

I don't feel as confident about this one. It felt like a good win when the name conflicted with `std::borrow::Borrow`, but renaming it to `TypedArray` resolves that issue. It doesn't feel super rusty to duplicate these.

We could add them later with a semver compatible change. What do you think about deferring it for now?

kjvalencik: The other change I think is to leave `BorrowError` named that instead of `TypedArrayError`. It's unclear what that error is and the error type is scoped to the `buffer` module.

kjvalencik:
Merged in https://github.com/neon-bindings/rfcs/commit/8beeb16a87446f32582b106b2716243798e6f89b

PR #45, opened 2021-11-11
Closure functions

kjvalencik:
Proposal to allow creating `JsFunction` from Rust `Fn` closures.

[Implementation](https://github.com/neon-bindings/neon/pull/811)

[Rendered](https://github.com/neon-bindings/rfcs/blob/kv/closure-functions/text/0000-closure-functions.md)

### DIFF
--- a/text/0000-closure-functions.md
+++ b/text/0000-closure-functions.md
@@ -168,4 +168,4 @@ Instead, we will only provide this feature for Node-API 5+. This costs a small a
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-* ~~Do closures need to be `Send`?~~ `V8` heavily uses thread local storage for the vent loop. It would be an extremely large change to be able to switch the thread running the event loop. It would also require changes to `libuv`. Even thought it's not _guaranteed_ this won't change, it seems unlikely enough that the small risk is worth the improved API.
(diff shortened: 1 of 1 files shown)


text/0000-closure-functions.md
```diff
@@ -0,0 +1,171 @@
...
+[alternatives]: #alternatives
+
+The primary alternative considered was to allow closures on *all* Node-API versions. However, it is very subtle that this would leak. Especially since the leak is determined by Neon feature flags and not based on the current runtime.
+
+Instead, we will only provide this feature for Node-API 5+. This costs a small amount of code complexity; however, we may be able to remove this later when all versions of Node on our support matrix provide Node-API 5.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* ~~Do closures need to be `Send`?~~ `V8` heavily uses thread local storage for the vent loop. It would be an extremely large change to be able to switch the thread running the event loop. It would also require changes to `libuv`. Even thought it's not _guaranteed_ this won't change, it seems unlikely enough that the small risk is worth the improved API.
```
dherman, line 171:
```suggestion
* ~~Do closures need to be `Send`?~~ `V8` heavily uses thread local storage for the event loop. It would be an extremely large change to be able to switch the thread running the event loop. It would also require changes to `libuv`. Even thought it's not _guaranteed_ this won't change, it seems unlikely enough that the small risk is worth the improved API.
```

dherman (approved):
This is such a great improvement to the `JsFunction` API, and the implementation really helped us work out the details.. Looks great to me!

kjvalencik:
Merged in https://github.com/neon-bindings/rfcs/commit/079dc4302903fac8d2fc4bb219aa7c72f5eb75c2

PR #46, opened 2022-02-14
Futures RFC

kjvalencik:


kjvalencik:
I propose that we only implement this for Node-API >= 5, allowing the usage of `JsFunction::new` from closures. This simplifies the state management.

If we would like to support this for lower versions, we will need to find an alternate way to track the state. For example, putting it in a `JsBox` and calling `Function.prototype.bind` on our `.then` and `.catch` methods.

PR #47, opened 2022-07-24
RFC: typed array conveniences

dherman:
This RFC defines a number of missing convenience APIs for typed arrays: constructors and metadata queries.

Implementation: neon-bindings/neon#909

[Rendered](https://github.com/neon-bindings/rfcs/blob/b400353cbbe41693094d5cda62b909e63482b919/text/0000-typed-array-conveniences.md)

### DIFF
--- a/text/0000-typed-array-conveniences.md
+++ b/text/0000-typed-array-conveniences.md
@@ -1,0 +1,312 @@
+- Feature Name: typed_array_conveniences
+- Start Date: 2022-07-23
+- RFC PR: (leave this empty)
+- Neon Issue: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Core support for reading and writing typed arrays was added in [RFC 39](https://github.com/neon-bindings/rfcs/blob/main/text/0039-node-api-borrow.md). This RFC adds APIs for conveniently **constructing and accessing metadata from** typed arrays.
+
+# Motivation
+[motivation]: #motivation
+
+Constructing typed arrays can be done in JavaScript or by accessing the constructors from `cx.global()`, but this requires a lot of boilerplate and is slower than calling the direct Node-API functions for doing the same.
+
+Reading metadata such as the byte length can be performed by borrowing a typed array's contents and doing arithmetic on the slice size. But this is error-prone and not particularly discoverable.
+
+Having a richer API for working with typed arrays also makes for a better learning experience in the API docs.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+This example creates a typed array of unsigned 32-bit integers with a user-specified length:
+
+```rust
+fn create_int_array(mut cx: FunctionContext) -> JsResult<JsTypedArray<u32>> {
+    let len = cx.argument::<JsNumber>(0)?.value(&mut cx) as usize;
+    JsTypedArray::new(&mut cx, len)
+}
+```
+
+This example creates a typed array as a view over a region of an existing buffer:
+
+```rust
+// Allocate a 16-byte ArrayBuffer and a uint32 array of length 2 (i.e., 8 bytes)
+// starting at byte offset 4 of the buffer:
+//
+//       0       4       8       12      16
+//      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// buf: | | | | | | | | | | | | | | | | |
+//      +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+//               ^       ^
+//               |       |
+//              +-------+-------+
+//         arr: |       |       |
+//              +-------+-------+
+//               0       1       2
+let buf = cx.array_buffer(16);
+let arr = JsUint32Array::from_region(&mut cx, buf.region(4, 2))?;
+```
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+## Convenience types
+
+The `neon::types` module contains the following convenient type aliases:
+
+| Rust Type           | Convenience Type   | JavaScript Type                    |
+| --------------------| ------------------ | ---------------------------------- |
+| `JsTypedArray<u8>`  | `JsUint8Array`     | [`Uint8Array`][Uint8Array]         |
+| `JsTypedArray<i8>`  | `JsInt8Array`      | [`Int8Array`][Int8Array]           |
+| `JsTypedArray<u16>` | `JsUint16Array`    | [`Uint16Array`][Uint16Array]       |
+| `JsTypedArray<i16>` | `JsInt16Array`     | [`Int16Array`][Int16Array]         |
+| `JsTypedArray<u32>` | `JsUint32Array`    | [`Uint32Array`][Uint32Array]       |
+| `JsTypedArray<i32>` | `JsInt32Array`     | [`Int32Array`][Int32Array]         |
+| `JsTypedArray<u64>` | `JsBigUint64Array` | [`BigUint64Array`][BigUint64Array] |
+| `JsTypedArray<i64>` | `JsBigInt64Array`  | [`BigInt64Array`][BigInt64Array]   |
+| `JsTypedArray<f32>` | `JsFloat32Array`   | [`Float32Array`][Float32Array]     |
+| `JsTypedArray<f64>` | `JsFloat64Array`   | [`Float64Array`][Float64Array]     |
+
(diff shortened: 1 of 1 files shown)


text/0000-typed-array-conveniences.md
```diff
@@ -0,0 +1,312 @@
...
+}
+```
+
+### `JsArrayBuffer`
+
+The `JsArrayBuffer` type gets a new method for constructing a `Region` struct, which describes a region of a buffer (see below).
+
+```rust
+impl Handle<'cx, JsArrayBuffer> {
+    pub fn region<T: Binary>(self, byte_offset: usize, len: usize) -> Region<'cx, T>;
```
kjvalencik: Should this be `to_region` for symmetry with the other methods? Also, I think `&self` is better unless there's a benefit of `self`.

text/0000-typed-array-conveniences.md
```diff
@@ -0,0 +1,312 @@
...
+}
+```
+
+### Helper types
+
+#### `neon::types::binary::Binary`
```
kjvalencik: I think it would be better to put `Binary` and `Region` in the existing `neon::types::buffer` module to keep the tree slightly smaller and to co-locate array buffer related traits/types.

https://docs.rs/neon/latest/neon/types/buffer/index.html

text/0000-typed-array-conveniences.md
```diff
@@ -0,0 +1,312 @@
...
+
+```rust
+#[derive(Clone, Copy)]
+struct Region<'cx, T: Binary> { /* hidden */ }
+
+impl<'cx, T: Binary> Region<'cx, T> {
+    // Constructs a new typed array for this region.
+    pub fn to_typed_array<'c, C>(self, cx: &mut C) -> JsResult<'c, JsTypedArray<T>>
+    where C: Context<'c>;
+}
```
kjvalencik, line 153: I think a nice addition would be to add a `to_region` on `Region` for nice chaining sub-regions and converting types.

```rust
impl<'cx, T: Binary> Region<'cx, T> {
    // Constructs a new typed array for this region.
    pub fn to_typed_array<'c, C>(&self, cx: &mut C) -> JsResult<'c, JsTypedArray<T>>
    where
        C: Context<'c>;

    // Constructs a new sub-region
    pub fn to_region<'c, U, C>(&self, byte_offset: usize, len: usize) -> Region<'cx, U>
    where
        C: Context<'c>,
        U: Binary;
}
```

```rust
let numbers = cx.argument::<JsUint8Array>(0)?
    .to_region(&mut cx)
    .to_region<JsUint32Array, _>(&mut cx, 2, 4)
    .to_typed_array(&mut cx)?;
```

text/0000-typed-array-conveniences.md
```diff
@@ -0,0 +1,312 @@
...
+
+The `Region` struct defines a region of an underlying buffer. For performance, it's not actually checked to be a valid region until actually constructing a typed array object. (Otherwise, every method on the type would have to re-validate every time it's called, because of the possibility of the buffer getting detached.)
+
+```rust
+#[derive(Clone, Copy)]
+struct Region<'cx, T: Binary> { /* hidden */ }
+
+impl<'cx, T: Binary> Region<'cx, T> {
+    // Constructs a new typed array for this region.
+    pub fn to_typed_array<'c, C>(self, cx: &mut C) -> JsResult<'c, JsTypedArray<T>>
```
kjvalencik:
```suggestion
    pub fn to_typed_array<'c, C>(&self, cx: &mut C) -> JsResult<'c, JsTypedArray<T>>
```

Methods prefix `to_` should take `&self`. Methods that take `self` should be named `into`. Since we're not really consuming anything, I think `&self` and keeping the method makes more sense.

text/0000-typed-array-conveniences.md
```diff
@@ -0,0 +1,312 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+There aren't any real drawbacks to exposing this functionality, since it's provided by Node-API. There are, however, a number of tradeoffs in the particulars of the design.
+
+The main drawbacks of this RFC's proposed design are:
+
+## Naming inconsistency: `len()` vs `byte_length()`
+
+This RFC goes with `len()` to match a strong Rust convention, and `byte_length()` to match the typed array naming convention (using snake case, of course). These does lead to an inconsistency between the two. A couple alternatives would be `byte_len()` or `size()`. But then `byte_len()` ends up feeling a bit inconsistent with `byte_offset()`. We could then choose `byte_off()` but that's a bit opaque.
```
kjvalencik: I somewhat prefer `byte_len` and `byte_offset` to `byte_length`. I agree that `byte_off` is confusing.

Alternatively, `byte_length` could be omitted completely since it's pretty easy to implement.

```
let byte_length = std::mem::size_of::<u32>() * arr.len(&mut cx);
```

The problem with this code is it's really easy for the generic in `size_of` to fall out of sync with the type of `T` in `arr`.

text/0000-typed-array-conveniences.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Unfortunate docs placement of `region()`
+
+In order to make the lifetimes work out, `region()` is a method of `Handle<JsArrayBuffer>` instead of `JsArrayBuffer`. This has the unfortunate effect of placing the method in the API docs for [`neon::handle::Handle`][neon::handle::Handle] instead of [`neon::types::JsArrayBuffer`][neon::types::JsArrayBuffer], which makes them less discoverable. The other docs should make an effort to mention the method prominently, use it in examples, and link to its docs in order to mitigate this issue.
```
kjvalencik, line 251: I don't intuitively understand the issue with lifetimes since in `&'a Handle<'cx, Uint8Array>` an `'a` could be chosen that lives as long as `'cx`.

text/0000-typed-array-conveniences.md
```diff
@@ -0,0 +1,312 @@
...
+
+## Flat constructor for `from_region`
+
+This RFC's design makes `from_region` and `to_region` symmetric signatures with `Region` as an input type vs. return type, respectively. An alternative approach would be to make `from_region` take the `byte_offset` and `len` as immediate arguments of `from_region`, and have `to_region` return a tuple. However, this would risk being more error-prone: users would have to remember what the two different `usize` elements of the tuple represent, and it would be easy to confuse the `len` integer for a byte length. It would also be untyped (see above).
+
+## Builder pattern for `Region`
+
+Instead of `buffer.region(offset, len)`, we could expose builder methods for constructing regions, like `buffer.byte_offset(offset).len(len)`. This would have the benefit of making the semantics of the integers more unambiguous, but doesn't seem worth the extra verbosity.
+
+## Alternative names for `Region`
```
kjvalencik, line 307: I really like `Region`.

kjvalencik (commented):
This design looks really great! I'm exited for the flexibility this provides.

I left some of my thoughts.

text/0000-typed-array-conveniences.md
dherman: What if went the other direction, and had them both called `.region()`? In both cases it's not really a conversion, it's a kind of accessor.

text/0000-typed-array-conveniences.md
dherman: Whoops, that was just a brain-o when I typed up the RFC (the implementation PR already agrees with your suggestion). I'll fix the doc.

text/0000-typed-array-conveniences.md
dherman: Makes sense. It's a bit of a theoretical distinction here since `Region` is `Copy`, but that might still be confusing from a documentation perspective, so I like your suggestion.

text/0000-typed-array-conveniences.md
dherman: I could go with `byte_len` and `byte_offset`. But then we just had a discussion and really liked the cleanness of:

```rust
offset()
len()
size()
```

text/0000-typed-array-conveniences.md
kjvalencik: This is to allow downcasting *and* returning a region in the same method. Otherwise, you would have a compiler error returning a reference to a dropped value.

text/0000-typed-array-conveniences.md
kjvalencik: I like `region`. Seeing it as an accessor (which should just be the name of the field, per API guidelines), this works perfectly.

kjvalencik (approved):
❤️ this design!

